PR #6, opened 2015-01-12
Remove `ORDER BY` statement

LukasReschke:
SQLite seems not to work here with a `ORDER BY` statement. - This prevented updates from previous versions.

@VicDeo Please review.

### DIFF
--- a/appinfo/preupdate.php
+++ b/appinfo/preupdate.php
@@ -5,7 +5,7 @@ if (version_compare($installedVersion, '0.6', '<')) {
 	$query = OC_DB::prepare( 'SELECT COUNT(*) AS `count`, `fileid` FROM `*PREFIX*files_antivirus` GROUP BY `fileid` HAVING COUNT(*) > 1' );
 	$result = $query->execute();
 	while( $row = $result->fetchRow()) {
-		$deleteQuery = OC_DB::prepare('DELETE FROM `*PREFIX*files_antivirus` WHERE `fileid` = ? ORDER BY `check_time` ASC', $row['count']-1);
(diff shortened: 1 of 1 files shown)


karlitschek:
:+1:

nickvergessen:
:+1:

appinfo/preupdate.php
```diff
@@ -5,7 +5,7 @@
 	$query = OC_DB::prepare( 'SELECT COUNT(*) AS `count`, `fileid` FROM `*PREFIX*files_antivirus` GROUP BY `fileid` HAVING COUNT(*) > 1' );
 	$result = $query->execute();
 	while( $row = $result->fetchRow()) {
-		$deleteQuery = OC_DB::prepare('DELETE FROM `*PREFIX*files_antivirus` WHERE `fileid` = ? ORDER BY `check_time` ASC', $row['count']-1);
+		$deleteQuery = OC_DB::prepare('DELETE FROM `*PREFIX*files_antivirus` WHERE `fileid` = ?', $row['count']-1);
```
nickvergessen: What kind of hack is `fileid = count -1` this makes NO sense at all, unless its explained in a comment.

LukasReschke: I certainly don't know. – I think it removes all entries except one… - There was no PK on this table before.

VicDeo: @LukasReschke correct. Adding a PK it appeared that table might have several scan results if the app is old enough. This cleans up scan results except the very last.

nickvergessen: Okay then

PR #8, opened 2015-01-12
Disable proxies

VicDeo:


DeepDiver1975:
Please add clover.xml to gitignore and do not check it in - THX

DeepDiver1975:
> Please add clover.xml to gitignore

Did you miss to add .gitignore ? THX

VicDeo:
@DeepDiver1975 well, I can make it pass now, but the faulty piece is https://github.com/owncloud/core/blob/master/apps/files_encryption/lib/proxy.php#L64

Not every path  has  `$parts[2]`. e.g.  `/foo.txt` hasn't

DeepDiver1975:
@schiesbn can you please have a look - THX

schiesbn:
Maybe we the proxy should check the number of path elements. But can you, @VicDeo, explain to me on what files the antivirus app operates to get a better understanding of the overall problem. The path at the proxy should always start at ```data/```, does the antivirus app operates on files outside of ```data/<user>/files``` ?

nickvergessen:
prepared a patch for it in owncloud/core#13285

VicDeo:
@schiesbn not exact, unit tests use temporary storage that is mounted to the user root directory

VicDeo:
@nickvergessen thanks! :icecream:

nickvergessen:
@VicDeo since my patch got merged, you just need to rebase it here.

VicDeo:
Tests are passing

Issue #38, opened 2015-04-17
files_antivirus app does not pass occ app:check-code


jnfrmarks:
### Steps to reproduce

1. Install 8.1 enterprise
2. invoke occ app:check-code

### Expected behaviour

The app passes the check

### Actual behaviour

The app fails the check:

```

[32mAnalysing /var/www/html/owncloud/apps/files_antivirus/tests/statustest.php[39m
 1 errors
    [37;41mline   23: OC_App - static method call on private class[39;49m
[32mAnalysing /var/www/html/owncloud/apps/files_antivirus/tests/itemtest.php[39m
 3 errors
    [37;41mline   24: OC_User - static method call on private class[39;49m
    [37;41mline   25: OC_User - static method call on private class[39;49m
    [37;41mline   29: OC_User - static method call on private class[39;49m
[32mAnalysing /var/www/html/owncloud/apps/files_antivirus/tests/testbase.php[39m
 1 errors
    [37;41mline   25: OC_App - static method call on private class[39;49m
[32mAnalysing /var/www/html/owncloud/apps/files_antivirus/tests/scannertest.php[39m
 2 errors
    [37;41mline   45: OC_User - static method call on private class[39;49m
    [37;41mline   49: OC_User - static method call on private class[39;49m
[32mAnalysing /var/www/html/owncloud/apps/files_antivirus/tests/bootstrap.php[39m
 1 errors
    [37;41mline   14: OC_Hook - static method call on private class[39;49m
[32mAnalysing /var/www/html/owncloud/apps/files_antivirus/templates/notification.php[39m
 0 errors
[32mAnalysing /var/www/html/owncloud/apps/files_antivirus/templates/settings.php[39m
 0 errors
[32mAnalysing /var/www/html/owncloud/apps/files_antivirus/controller/settingscontroller.php[39m
 0 errors
[32mAnalysing /var/www/html/owncloud/apps/files_antivirus/controller/rulecontroller.php[39m
 0 errors
[32mAnalysing /var/www/html/owncloud/apps/files_antivirus/lib/item.php[39m
 2 errors
    [37;41mline  185: OC_DB - static method call on private class[39;49m
    [37;41mline  190: OC_DB - static method call on private class[39;49m
[32mAnalysing /var/www/html/owncloud/apps/files_antivirus/lib/notification.php[39m
 0 errors
[32mAnalysing /var/www/html/owncloud/apps/files_antivirus/lib/backgroundscanner.php[39m
 4 errors
[...]

PR #39, opened 2015-04-20
Fix some code checker complaints

VicDeo:
Ref https://github.com/owncloud/files_antivirus/issues/38

### DIFF
--- a/tests/statustest.php
+++ b/tests/statustest.php
@@ -20,7 +20,6 @@ class Test_Files_Antivirus_Status extends  \OCA\Files_Antivirus\Tests\Testbase {
 
 	public function setUp() {
 		parent::setUp();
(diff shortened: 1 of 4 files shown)


tests/statustest.php
```diff
@@ -20,7 +20,6 @@ class Test_Files_Antivirus_Status extends  \OCA\Files_Antivirus\Tests\Testbase {
 
 	public function setUp() {
 		parent::setUp();
-		\OC_App::enable('files_antivirus');
```
nickvergessen: you can keep this now if required, because tests/ is now ignored by the code checker

nickvergessen:
:+1:

tests/statustest.php
VicDeo: @nickvergessen this one is left in a base class :)

Issue #66, opened 2015-08-05
Change recipient name in notification mail if using user_ldap


RostC:
If I use files_antivirus and user_ldap app the recipient name in the notification mail is in standard config something like 
D14AB3F9-08F1-43AE-B83D-1D678A8A5416 <mail@myhost.com>

This is because the "username" in the database is  the AD objectGUID.
Would be great if in this case the displayname for example is the recipient name.

karlitschek:
@VicDeo What do you think?

PR #67, opened 2015-08-05
change recipient name if user_ldap is used

RostC:
fixes: #66 
Possible solution to check if user_ldap is used and use then an other attribute as recipient name

### DIFF
--- a/lib/notification.php
+++ b/lib/notification.php
@@ -14,6 +14,15 @@ class Notification {
 			return;
 		}
 		$email = \OCP\Config::getUserValue(\OCP\User::getUser(), 'settings', 'email', '');
+		if (\OCP\App::isEnabled(user_ldap)){
+                        $user  = \OCP\Config::getUserValue(\OCP\User::getUser(), 'user_ldap', 'displayName', '');
+                        if (empty($user)) {
+                                $user = \OCP\User::getUser();
+                        }
+                }
+                else {
+                        $user = \OCP\User::getUser();
+                };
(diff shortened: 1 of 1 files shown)


lib/notification.php
```diff
@@ -14,6 +14,15 @@ public static function sendMail($path){
 			return;
 		}
 		$email = \OCP\Config::getUserValue(\OCP\User::getUser(), 'settings', 'email', '');
+		if (\OCP\App::isEnabled(user_ldap)){
```
DeepDiver1975: Hmm .... why not do this in any case? @VicDeo objections

DeepDiver1975: We might even just display both user id and displayname - since both have a reasonable meaning

RostC: we can do this without the check but I thought It would be better to check first then have a statement $user  = \OCP\Config::getUserValue(\OCP\User::getUser(), 'user_ldap', 'displayName', ''); which is always empty (aka fails) if user_ldap is not enabled. But I agree the behavior is the same

RostC: I think it depends on what's the correct value in database. It is not helpful (in my opinion) if the mailheader looks like 
D14AB3F9-08F1-43AE-B83D-1D678A8A5416 <mail@myhost.com>
should be
My Name <mail@myhost.com>

DeepDiver1975:
@VicDeo can I ask you to have a look at this - THX

VicDeo:
@DeepDiver1975 technically correct, but the whole class needs to be updated in order not to use deprecated methods
I think it's good to go for now :+1:

Issue #69, opened 2015-08-06
"fwrite(): send of 4 bytes failed with errno=32 Broken pipe at \/var\/www\/owncloud\/apps\/files_antivirus\/lib\/scanner.php#68


Daniae:
Hi,

i have a Problem with ownCloud 8.1.0 and files_antivirus 0.7.0.1. I got the following message several times in a second in the owncloud.log:

remoteAddr":"","app":"PHP","message":"fwrite(): send of 4 bytes failed with errno=32 Broken pipe at \/var\/www\/owncloud\/apps\/files_antivirus\/lib\/scanner.php#68","level":3,

I can not find the source of this problem. Is there anyone who knows why this message appears and how i can fix this problem?

DeepDiver1975:
@VicDeo please have a look - THX

Daniae:
Hi,

i found the cause of this message. The Problem appears if the cron.php from owncloud runs the first time on the server. If i start the Server without using the cron.php, i get no entry in the owncloud.log. I use the Cron entry described in the ownlcoud 8.1 documentation: https://doc.owncloud.org/server/8.1/admin_manual/configuration_server/background_jobs_configuration.html?highlight=cron

I hope these information help you find the source of the Problem. Thank you for the help!

VicDeo:
@Daniae 
What exact cron mode you are using (webcron, ajax. etc)?
What are the files_antivirus app settings?

Daniae:
Hi VicDeo,

i use cron with the cron.php. My Crontab has this entry: 

*/15  *  *  *  * php -f /var/www/owncloud/cron.php > /dev/null 2>&1

This are my settings in the administration panel for the files_antivirus app:

Modus: Deamon Socket
Socket: /var/run/clamav/clamd.ctl
Übertragunsgslänge: 10485760
Aktion: Loggen

This is my /etc/clamav/clamd.conf:

#Automatically Generated by clamav-daemon postinst
#To reconfigure clamd run #dpkg-reconfigure clamav-daemon
#Please read /usr/share/doc/clamav-daemon/README.Debian.gz for details
LocalSocket /var/run/clamav/clamd.ctl
FixStaleSocket true
LocalSocketGroup clamav
LocalSocketMode 666
# TemporaryDirectory is not set to its default /tmp here to make overriding
# the default with environment variables TMPDIR/TMP/TEMP possible
User clamav
AllowSupplementaryGroups false
ScanMail true
ScanArchive true
[...]

PR #82, opened 2015-11-17
Skip zero sized files

VicDeo:
Fixes https://github.com/owncloud/files_antivirus/issues/69


karlitschek:
:+1:

DeepDiver1975:
:+1:

fsbruva:
This does not actually fix the problem. If there are 5 or more zero sized files, the background scanner will continuously select them from the database, but skip them in the while loop.

I will be submitting a pull request shortly that correctly fixes the problem.

fsbruva:
Sorry for the multiple references listed. I had some trouble with my forked repository, and ended up with (obviously) several duplicate commits. The a1a3116 is the correct one, referenced in my pull request.

PR #83, opened 2015-11-18
Bugfix for #82, and fixes #69 

### LINKED ISSUE #69: "fwrite(): send of 4 bytes failed with errno=32 Broken pipe at \/var\/www\/owncloud\/apps\/files_antivirus\/lib\/scanner.php#68

fsbruva:
The proposed patch (#82) for bug #69 will repeatedly (every cronjob) select the empty files from the filecache, and then skip over them. Their fileid's will never be added to the files_antivirus table, so they will always be selected. Once there are more than 5 empty files, the background scanner will select those 5 as a result of the JOIN and LEFT JOIN, and will skip them all. This will leave the remainder of the files on the server unprotected.

A better solution is to only consider files with non-zero sizes in the first place.

### DIFF
--- a/lib/backgroundscanner.php
+++ b/lib/backgroundscanner.php
@@ -59,7 +59,8 @@ class BackgroundScanner {
 			.' WHERE `mimetype` != ?'
 			.' AND (`*PREFIX*storages`.`id` LIKE ? OR `*PREFIX*storages`.`id` LIKE ?)'
 			.' AND (`*PREFIX*files_antivirus`.`fileid` IS NULL OR `mtime` > `check_time`)'
-			.' AND `path` LIKE ?';
+			.' AND `path` LIKE ?'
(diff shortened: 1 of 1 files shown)


VicDeo:
@fsbruva thanks!
:+1:

fsbruva:
@VicDeo No worries! Sorry I cluttered things up with my commits. I hated having "Charlie Root" as the committer.

PR #111, opened 2016-05-04
 Antivirus update

VicDeo:
- Where possible: Use `===`, remove deprecated calls 
- Fix saving rules at admin page
- Fix broken tests

### DIFF
--- a/lib/backgroundscanner.php
+++ b/lib/backgroundscanner.php
@@ -82,11 +82,11 @@ class BackgroundScanner {
 					$qb->expr()->like('fc.path', $qb->expr()->literal('files/%'))
 				)
 				->andWhere(
-					$qb->expr()->neq('fc.size', '?1')
+					$qb->expr()->neq('fc.size', '0')
 				)
 				->setMaxResults(5)
 			;
-			$result = $qb->execute([0]);
+			$result = $qb->execute();
 		} catch(\Exception $e) {
 			\OC::$server->getLogger()->error( __METHOD__ . ', exception: ' . $e->getMessage(), ['app' => 'files_antivirus']);
 			return;
--- a/lib/backgroundscanner.php
+++ b/lib/backgroundscanner.php
@@ -51,50 +51,76 @@ class BackgroundScanner {
 			return;
 		}
 		// locate files that are not checked yet
-		$dirMimetypeId = \OC::$server->getMimeTypeLoader()->getId('httpd/unix-directory');
-		$sql = 'SELECT `*PREFIX*filecache`.`fileid`, `*PREFIX*storages`.*'
-			.' FROM `*PREFIX*filecache`'
-			.' LEFT JOIN `*PREFIX*files_antivirus` ON `*PREFIX*files_antivirus`.`fileid` = `*PREFIX*filecache`.`fileid`'
-			.' JOIN `*PREFIX*storages` ON `*PREFIX*storages`.`numeric_id` = `*PREFIX*filecache`.`storage`'
-			.' WHERE `mimetype` != ?'
-			.' AND (`*PREFIX*storages`.`id` LIKE ? OR `*PREFIX*storages`.`id` LIKE ?)'
-			.' AND (`*PREFIX*files_antivirus`.`fileid` IS NULL OR `mtime` > `check_time`)'
-			.' AND `path` LIKE ?'
-			.' AND `*PREFIX*filecache`.`size` != 0';
-		$stmt = \OCP\DB::prepare($sql, 5);
+		$dirMimeTypeId = \OC::$server->getMimeTypeLoader()->getId('httpd/unix-directory');
 		try {
-			$result = $stmt->execute(array($dirMimetypeId, 'local::%', 'home::%', 'files/%'));
-			if (\OCP\DB::isError($result)) {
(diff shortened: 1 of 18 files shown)


VicDeo:
@icewind1991 may I ask  you to review this PR?

lib/backgroundscanner.php
```diff
@@ -51,58 +51,83 @@ public function run(){
...
+					$qb->expr()->orX(
+						$qb->expr()->isNull('fa.fileid'),
+						$qb->expr()->gt('fc.mtime', 'fa.check_time')
+					)
+				)
+				->andWhere(
+					$qb->expr()->like('fc.path', $qb->expr()->literal('files/%'))
+				)
+				->andWhere(
+					$qb->expr()->neq('fc.size', '?1')
```
icewind1991: `'?1'` should be `0`?

lib/notification.php
```diff
@@ -13,34 +13,36 @@ public static function sendMail($path){
...
-                                $user = \OCP\User::getUser();
-                        }
-                }
-                else {
-                        $user = \OCP\User::getUser();
-                };
+			$displayName = $config->getUserValue($user->getUID(), 'user_ldap', 'displayName', '');
+		}
+		if (empty($displayName)) {
+			$displayName = $user->getUID();
```
icewind1991: same here

icewind1991:
Looks good otherwise

lib/notification.php
VicDeo: `return null` in `User_LDAP::getDisplayName` https://github.com/owncloud/core/blob/master/apps/user_ldap/lib/User_LDAP.php#L386 thinks differently

icewind1991: `$user->getDisplayName` does handle it though

PVince81:
@VicDeo from the IRC discussion it seems that this fix is critical to make antivirus work again with OC >= 9.0 ?

PVince81:
@VicDeo how to test ? Is this only about background scanning, not upload time scanning ?

CC @owncloud/qa

PVince81:
Will need backport to 9.0 and 9.1

CC @DeepDiver1975 @dragotin

VicDeo:
@PVince81 it doesn't fix backround job itself, but a couple of other issues (see the first message)
But the background job is changed with this PR so it's better to merge it first

Fix saving rules at admin page
1. modify rule in advanced section at the admin page
2. click save
3. refresh page

Expected: changes are saved
Actual: they are not

Also it moves adding a cron task to the install script instead of adding it any time app.php is executed.

SergioBertolinSG:
Doesn't work for me. :-1:

The rule is not saved.

Also if the save button is required also for rules, it should be duplicated or moved to the bottom of the antivirus section. Otherwise it is a bad UX.
 
BTW what is the difference in rules between "Infected" and "Clean" ?

VicDeo:
@SergioBertolinSG 
>save button is required also for rules

It is not. It saves non-advanced section.
As soon as rule was added, it is saved by clicking checkmark at the left. The same is true for changing a rule

>BTW what is the difference in rules between "Infected" and "Clean" ?

If av binary/socket/etc response contains something that matches "Infected" rule, the file is considered to be infected.
As soon all infected rules are tested, the response is checked against "Clean" rules.

In no rules matched the file status is unknown.

SergioBertolinSG:
OK, I didn't understand the interface. I suggest an icon change, something like ✅  for me doesn't look like a save button. 

:+1:

lib/backgroundscanner.php
```diff
@@ -51,58 +51,83 @@ public function run(){
...
+				->leftJoin('fc', 'files_antivirus', 'fa', $qb->expr()->eq('fa.fileid', 'fc.fileid'))
+				->innerJoin(
+					'fc',
+					'storages',
+					'ss',
+					$qb->expr()->andX(
+						$qb->expr()->eq('fc.storage', 'ss.numeric_id'),
+						$qb->expr()->orX(
+							$qb->expr()->like('ss.id', $qb->expr()->literal('local::%')),
+							$qb->expr()->like('ss.id', $qb->expr()->literal('home::%'))
```
PVince81: note: this won't work for md5 storage ids https://github.com/owncloud/core/wiki/Storage-IDs#random-numberstring
Will need separate fixing.

The way how repair steps do is iterate over all users and get their home storages... yes, cumbersome.

PVince81: Ah, and starting with OC 9.0 it will be possible to found the home mount from the oc_mounts table

lib/backgroundscanner.php
```diff
@@ -51,58 +51,83 @@ public function run(){
...
+					$qb->expr()->neq('fc.mimetype', $qb->expr()->literal($dirMimeTypeId))
+				)
+				->andWhere(
+					$qb->expr()->orX(
+						$qb->expr()->isNull('fa.fileid'),
+						$qb->expr()->gt('fc.mtime', 'fa.check_time')
+					)
+				)
+				->andWhere(
+					$qb->expr()->like('fc.path', $qb->expr()->literal('files/%'))
```
PVince81: Assuming this never runs on external storage.

PVince81:
Code looks good :+1: 

Needs backport ?

VicDeo:
@PVince81 yes, otherwise rules in advanced section are not changed/deleted

PR #113, opened 2016-05-12
Well .. opps I guess?

nickvergessen:
@VicDeo please take care of backporting to 8.2 and 9.0


VicDeo:
@nickvergessen thanks! :+1:

PVince81:
:+1:

nickvergessen:
Okay I just tested and it "works" as intended but creates a notice: https://3v4l.org/CK8er

So good to fix, but shouldn't break functionality as I thought yesterday

Issue #109, opened 2016-04-24
Backgroundscan not working - Error Failed to setup file system in backgroundscanner.php


marco-aa:
Hello all :-)

i'm using OwnCloud 8.2.3 and antivirus app 0.8.0.1. on Centos 7.2 running SELinux active. (checked audit2allow - looking good )

I found "Error  files_antivirus Failed to setup file system" in my Administrator Log.

I looked at backgroundscanner.php and i found that 
A)
in funciton initFS() the line $results = $this->userManager->search('', 2, 0); has no result.
userManager object is working. No result found. 
If i change  the line \OC_Util::setupFS($anyUser->getUID()); to \OC_Util::setupFS('my-admin-user');
the background scan is working. (i did a test with a EICA-Test-File).
B)
File is not deleted with my "workaround" (i checked my settings on the adminpanel delete and not logging was set)
C)
By the way will the setupFS only work an files under my "my-admin-user" or will it scan all files for example from the private folders of other owncloud users ?

Somebody has a idea ? Thanks a lot

goopen:
Not sure if this is related but I get this error in my logs. This is on 9.0.2 however.

```
[...]

PR #118, opened 2016-07-12
Resolve path by fileId correctly

### LINKED ISSUE #98: OC 9 - files_antivirus stable 9
Hey guys, next to the integrity check the logs also show this error:

```Error while running background job (class: OC\BackgroundJob\Legacy\RegularJob, arguments: Array ( [0] => OCA\Files_Antivirus\Cron\Task [1] => run ) ): {"Exception":"OCP\\Files\\NotFoundException","Message":"File with id \"4\" has not been found.","Code":0,"Trace":"#0 \/var\/www\/owncloud\/apps\/files_antivirus\/lib\/backgroundscanner.php(78): OC\\Files\\View->getPath(4)\n#1 \/var\/www\/owncloud\/apps\/files_antivirus\/lib\/cron\/task.php(21): OCA\\Files_Antivirus\\BackgroundScanner->run()\n#2 [internal function]: OCA\\Files_Antivirus\\Cron\\Task::run()\n#3 \/var\/www\/owncloud\/lib\/private\/backgroundjob\/legacy\/regularjob.php(28): call_user_func(Array)\n#4 \/var\/www\/owncloud\/lib\/private\/backgroundjob\/job.php(52): OC\\BackgroundJob\\Legacy\\RegularJob->run(Array)\n#5 \/var\/www\/owncloud\/cron.php(141): OC\\BackgroundJob\\Job->execute(Object(OC\\BackgroundJob\\JobList), Object(OC\\Log))\n#6 {main}","File":"\/var\/www\/owncloud\/lib\/private\/files\/view.php","Line":1639}```

Ubuntu 14.04
OC 9.0.0
PHP 7.0
Apache 2.4.7
MySql 5.6

Anyone?

### PR DESCRIPTION
**Pull request by VicDeo**
Fixes
#98 
https://github.com/owncloud/files_antivirus/issues/109
https://github.com/owncloud/core/issues/25759

### DIFF
--- a/lib/backgroundscanner.php
+++ b/lib/backgroundscanner.php
@@ -107,7 +107,7 @@ class BackgroundScanner {
 		}
 
 		$cnt = 0;
-		while ($row = $result->fetch() && $cnt < self::BATCH_SIZE) {
+		while (($row = $result->fetch()) && $cnt < self::BATCH_SIZE) {
 			try {
 				$fileId = $row['fileid'];
 				$owner = $this->getOwner($fileId);
--- a/lib/backgroundscanner.php
+++ b/lib/backgroundscanner.php
@@ -8,38 +8,57 @@
 
 namespace OCA\Files_Antivirus;
 
-use OCP\IUserManager;
+use OC\Files\Filesystem;
+use OC\Files\View;
 use OCP\IL10N;
-
-use OCA\Files_Antivirus\Item;
+use OCP\IUser;
+use OCP\Files\IRootFolder;
+use OCP\IUserSession;
 
 class BackgroundScanner {
+
+	const BATCH_SIZE = 10;
+
+	/** @var IRootFolder */
+	protected $rootFolder;
+
+	/** @var \OCP\Files\Folder[] */
+	protected $userFolders;
 
 	/**
 	 * @var ScannerFactory
 	 */
 	private $scannerFactory;
-	
-	/**
-	 * @var IUserManager 
-	 */
-	private $userManager;
+
 	
 	/**
 	 * @var IL10N
 	 */
 	private $l10n;
-	
+
+	/** @var string */
+	protected $currentFilesystemUser;
+
+	/** @var \OCP\IUserSession */
+	protected $userSession;
+
 	/**
 	 * A constructor
+	 *
 	 * @param \OCA\Files_Antivirus\ScannerFactory $scannerFactory
-	 * @param IUserManager $userManager
 	 * @param IL10N $l10n
+	 * @param IRootFolder $rootFolder
+	 * @param IUserSession $userSession
 	 */
-	public function __construct(ScannerFactory $scannerFactory, IUserManager $userManager, IL10N $l10n){
+	public function __construct(ScannerFactory $scannerFactory,
+								IL10N $l10n,
+								IRootFolder $rootFolder,
+								IUserSession $userSession
+	){
+		$this->rootFolder = $rootFolder;
 		$this->scannerFactory = $scannerFactory;
-		$this->userManager = $userManager;
 		$this->l10n = $l10n;
+		$this->userSession = $userSession;
 	}
 	
 	/**
@@ -47,9 +66,6 @@ class BackgroundScanner {
 	 * @return null
 	 */
 	public function run(){
-		if (!$this->initFS()) {
-			return;
-		}
 		// locate files that are not checked yet
 		$dirMimeTypeId = \OC::$server->getMimeTypeLoader()->getId('httpd/unix-directory');
 		try {
@@ -84,7 +100,7 @@ class BackgroundScanner {
 				->andWhere(
 					$qb->expr()->neq('fc.size', $qb->expr()->literal('0'))
 				)
-				->setMaxResults(5)
+				->setMaxResults(self::BATCH_SIZE)
 			;
 			$result = $qb->execute();
 		} catch(\Exception $e) {
@@ -92,13 +108,18 @@ class BackgroundScanner {
 			return;
 		}
 
-
-		$view = new \OC\Files\View('');
 		try {
 			while ($row = $result->fetch()) {
-				$path = $view->getPath($row['fileid']);
+				$fileId = $row['fileid'];
+				$owner = $this->getOwner($fileId);
+				if (!$owner){
+					continue;
+				}
+				$this->initFilesystemForUser($owner);
+				$view = \OC\Files\Filesystem::getView();
+				$path = $view->getPath($fileId);
 				if (!is_null($path)) {
-					$item = new Item($this->l10n, $view, $path, $row['fileid']);
+					$item = new Item($this->l10n, $view, $path, $fileId);
 					$scanner = $this->scannerFactory->getScanner();
 					$status = $scanner->scan($item);
 					$status->dispatch($item, true);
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lib/backgroundscanner.php [hunk 1]
```diff
@@ -84,21 +100,27 @@ public function run(){
...
 			return;
 		}
 
-
-		$view = new \OC\Files\View('');
 		try {
 			while ($row = $result->fetch()) {
-				$path = $view->getPath($row['fileid']);
+				$fileId = $row['fileid'];
+				$owner = $this->getOwner($fileId);
```
PVince81: Are you excluding federated shares here ? Just asking because today I noticed that it would get a non-resolvable owner (the federated id of the remote user).

lib/backgroundscanner.php
```diff
@@ -110,22 +132,42 @@ public function run(){
...
 	 */
-	protected function initFS(){
-		//Need any valid user to mount FS
-		$results = $this->userManager->search('', 2, 0);
-		$anyUser = array_pop($results);
-		if (is_null($anyUser)) {
-			\OC::$server->getLogger()->error("Failed to setup file system", ['app' => 'files_antivirus']);
-			return false;
+	protected function getUserFolder(IUser $user) {
+		if (!isset($this->userFolders[$user->getUID()])) {
```
PVince81: so these aren't cached in core already ?

PVince81:
What's left to do ?

I saw your bullet list here https://github.com/owncloud/files_antivirus/issues/109#issuecomment-230782092. Are these things we had before and need to re-add to this approach/PR ? Or does the PR match the old behavior feature-wise ?

lib/backgroundscanner.php
VicDeo: @PVince81 IDK, blindly copied from tag retention

lib/backgroundscanner.php
VicDeo: @PVince81 no. According to the query above :
```
 							$qb->expr()->like('ss.id', $qb->expr()->literal('local::%')),
 							$qb->expr()->like('ss.id', $qb->expr()->literal('home::%'))
```

VicDeo:
@PVince81 Are these things we had before and need to re-add to this approach/PR ? Or does the PR match the old behavior feature-wise ?
The latter (match the old behavior)

VicDeo:
Right now it produces the following:
```
PHP Fatal error:  Call to a member function getStorage() on a non-object in lib/private/Files/View.php on line 1688
PHP Stack trace:
PHP   1. {main}() cron.php:0
PHP   2. OC\BackgroundJob\TimedJob->execute() cron.php:122
PHP   3. OC\BackgroundJob\Job->execute() lib/private/BackgroundJob/TimedJob.php:53
PHP   4. OCA\Files_Antivirus\Cron\Task->run() lib/private/BackgroundJob/Job.php:52
PHP   5. OCA\Files_Antivirus\BackgroundScanner->run() apps/files_antivirus/lib/cron/task.php:32
PHP   6. OC\Files\View->getPath() apps/files_antivirus/lib/backgroundscanner.php:121
```

ms270169:
In _OC\Files\View.getPath()_:
line _$mounts = $manager->findIn($this->fakeRoot);_ is finding the needed mount
line _$mounts[] = $manager->find($this->fakeRoot);_ adds null to the array

After _array_reverse()_ the null is on first position. In the following foreach loop every array element is accessed, also the null. This leads to the fatal error.

Without this error (I prevent acess to null with an is_null(...)), it works.
Please don't forget some additional improvements on antivirus scan (see Conclusion in [issue #25759](https://github.com/owncloud/core/issues/25759#issuecomment-239664445) and [issue #125](https://github.com/owncloud/files_antivirus/issues/125))

lib/backgroundscanner.php
PVince81: ouch, this will not work with md5-style storage ids

PVince81:
@VicDeo I think the first goal is to achieve the same functionality like before first in this PR. After merging we can look into improving more stuff.

VicDeo:
@ms270169 thanks for debugging :beer: 
please retest, it should work now

VicDeo:
@PVince81 @DeepDiver1975 ready for review

lib/backgroundscanner.php
```diff
@@ -110,22 +131,43 @@ public function run(){
...
+	/**
+	 * @param IUser $user
+	 */
+	protected function initFilesystemForUser(IUser $user) {
+		if ($this->currentFilesystemUser !== $user->getUID()) {
+			if ($this->currentFilesystemUser !== '') {
+				Filesystem::tearDown();
+			}
+			Filesystem::init($user->getUID(), '/' . $user->getUID() . '/files');
+			$this->userSession->setUser($user);
```
PVince81: at you setting this back to null at some point ? Because multiple bkg jobs could run in a single pass so we don't want a stray user when running the next job...

PVince81:
@ms270169 did you get a chance to test this ? (I don't have the env setup to test it myself)

ms270169:
I have tested last commit c42ebb7b3c20ae9ba3a80d3e57d10a85d381957d.
Works as desired. All files were scanned succuessful.

Be aware, that there is some weakness in implementation.
For example, if a file cannot be found in file-system, scanning breaks down completly because _new Item_ will cause an exception, which breaks the loop.

PVince81:
> For example, if a file cannot be found in file-system, scanning breaks down completly because new Item will cause an exception, which breaks the loop.

@VicDeo Any chance to catch this and continue the loop ?

ms270169:
My recommendation: remove size limit in sql query and add inner try/catch with counter.
For example: 
```
try {
			$cnt = 0;
			while (($row = $result->fetch()) && $cnt++ < self::BATCH_SIZE) {
				$fileId = $row['fileid'];
				try {
					$owner = $this->getOwner($fileId);
					if (!$owner){
						continue;
					}
					$this->initFilesystemForUser($owner);
					$view = \OC\Files\Filesystem::getView();
					$path = $view->getPath($fileId);
					if (!is_null($path)) {
						$item = new Item($this->l10n, $view, $path, $fileId);
						$scanner = $this->scannerFactory->getScanner();
						$status = $scanner->scan($item);
						$status->dispatch($item, true);
					}
				} catch (\Exception $e){
					\OC::$server->getLogger()->warning( __METHOD__ . ', cannot scan fileid ' . $fileId . ', exception: ' . $e->getMessage(), ['app' => 'files_antivirus']);
				}
			}
		} catch (\Exception $e){
			\OC::$server->getLogger()->error( __METHOD__ . ', exception: ' . $e->getMessage(), ['app' => 'files_antivirus']);
		}
```
In case of missing file you will get two entries in log (error _File does not exist_ and warning _cannot scan file_). But every turn 10 files are scanned until all files are scanned.

VicDeo:
@ms270169 I disagree with 'remove size limit in sql query'. It filters  zero-sized files  that are definitely not infected and are known to produce errors as per https://github.com/owncloud/files_antivirus/issues/69

VicDeo:
@ms270169 
I got it. This totally makes sense.

@PVince81 try-catch moved into the loop, user-session reset to null and to prevent looping through unresolvable entries only successfully scanned filed are counted in the batch now

dercorn:
@RealRancor could this be backported to 9.0?

lib/backgroundscanner.php
```diff
@@ -84,48 +99,84 @@ public function run(){
...
-		$view = new \OC\Files\View('');
-		try {
-			while ($row = $result->fetch()) {
-				$path = $view->getPath($row['fileid']);
+		$cnt = 0;
+		while ($row = $result->fetch() && $cnt < self::BATCH_SIZE) {
+			try {
+				$fileId = $row['fileid'];
+				$owner = $this->getOwner($fileId);
+				if (!$owner){
```
PVince81: this might be true if `$owner` is equal to the string 0 ?

PVince81:
@VicDeo code looks good, I haven't tested this. :+1: 

If @ms270169 confirms that it still works then we can count it as a second thumbs up and merge.

PVince81:
@VicDeo how risky would be a backport to stable9 ?

VicDeo:
@PVince81 IMO backport is possible.

VicDeo:
Stable9.1:  https://github.com/owncloud/files_antivirus/issues/127
Stable9: https://github.com/owncloud/files_antivirus/issues/128

RealRancor:
Should also fix https://github.com/owncloud/files_antivirus/issues/98 or am i wrong?

@dercorn I have no idea if stuff can be backported :smile:

VicDeo:
@RealRancor Looks like this

ms270169:
Take attention to brackets!!
In [backgroundscanner.php line 110] (https://github.com/owncloud/files_antivirus/blob/89d4e7b17d1ed3dbad2279d43d4cb873f81bfaad/lib/backgroundscanner.php#L110):
`_while ($row = $result->fetch() && $cnt < self::BATCH_SIZE) {_`
.. row gets a boolean value instead of one row from query result.
Change it to
`_while (($row = $result->fetch()) && $cnt < self::BATCH_SIZE) {_`

VicDeo:
@ms270169 ouch... Thanks for spotting it!

VicDeo:
Stable9: https://github.com/owncloud/files_antivirus/pull/128
Stable9.1: https://github.com/owncloud/files_antivirus/pull/127

PVince81:
@ms270169 mind retesting ?

ms270169:
Everything works as desired.

I have tested cff2a478e4d71b037b8b1fe72f0a7849ce41912a with a setup where one file was deleted manually from file-system, but stays inside oc_filecache. So system is in an abnormal inconsistent state.
All valid files are scanned successfully. Two Exceptions are logged as error (level 3), caused by the missing file:

1. _"message":"File does not exist. File: 155 Path: ..."_
2. _"message":"OCA\\Files_Antivirus\\BackgroundScanner::run, exception:"_

Second exception my lead to confusion, it is thrown in [backgroundscanner.php#L130](https://github.com/owncloud/files_antivirus/blob/6bcff02985f1721d6ddd582d28edb2392a616c49/lib/backgroundscanner.php#L130) but _ex.getMessage()_ is empty. Adding some additional message text (like _"scanning file ... fails"_ would be helpful.

PVince81:
@VicDeo can you fix the message separately ?

Merging now so this can finally get into the release

PVince81:
@VicDeo does it need a backport to 8.2 ?

Issue #124, opened 2016-08-03
fwrite(): send of 1028 bytes failed with errno=32 Broken pipe at ....


alphamat:
I use cloud server 9 with debian 8 & php 7, apache 2.4, antivirus app daemon socket - also in use latest branche from here - i have in log a lot of this error messages:

`fwrite(): send of 4 bytes failed with errno=32 Broken pipe at /var/www/virtual/htdocs/apps/files_antivirus/lib/scanner/external.php#51

fwrite(): send of 931 bytes failed with errno=32 Broken pipe at /var/www/virtual/htdocs/apps/files_antivirus/lib/scanner.php#68

fwrite(): send of 1028 bytes failed with errno=32 Broken pipe at /var/www/virtual/htdocs/apps/files_antivirus/lib/scanner.php#68`

install & activation of the app was without any problem - i hope somebody can help with this issue

thx

Bliepo2:
To me this seems to be a bug. I have the exact same issue on my ownCloud, with ownCloud server 9.1, Debian Jessie. I also found similair reports at https://www.bountysource.com/issues/36929179-antivirus-data-stream-problem-fills-log-file.

Basseuph:
We face the same issue since the last update of owncloud and now deactivated virus scanning, which definitely is not the nice way of handling this. We run owncloud 9.1 (9.1.1-1.1 from owncloud repo) on Debian Jessie 8.6.

e-alfred:
Maybe related to this one?

https://github.com/owncloud/files_antivirus/issues/110

Bliepo2:
No, I don't think they are related. The only error I have is this one:
"Error 	PHP 	fwrite(): send of 4 bytes failed with errno=32 Broken pipe at /var/www/owncloud/apps/files_antivirus/lib/scanner/external.php#51 	2016-09-29T22:15:04+00:00
Error	PHP	fwrite(): send of 4 bytes failed with errno=32 Broken pipe at /var/www/owncloud/apps/files_antivirus/lib/scanner/external.php#51	2016-09-29T07:30:03+00:00
Error	PHP	fwrite(): send of 4 bytes failed with errno=32 Broken pipe at /var/www/owncloud/apps/files_antivirus/lib/scanner/external.php#51	2016-09-29T06:00:04+00:00
[...]

Issue #125, opened 2016-08-14
Antivirus data stream problem fills log file


ms270169:
In _OCA\Files_Antivirus\Scanner.initScanner()_ the nINSTREAM method is prepared to scan for a virus. In the clamav config file (_/etc/clamav/clamd.conf_) you can configure the maximum stream size with the line _StreamMaxLength 4095M_
With the value 4095M the maximum value of 4GiB is used, a bigger value is not supported by clamav.

Now in the function [OCA\Files_Antivirus\Scanner.scan(IScannable $item)](https://github.com/owncloud/files_antivirus/blob/master/lib/scanner.php#L61)  the scanning is done by cutting the file in small chunks (chunk size configureable, default size is 1024 bytes). Each chunk is streamed to the clamav-daemon.

If you have an video file with 5 GiB, there are 5368709120 chunks used. For the first 4294967296 chunks everything works fine.

But then the streaming size limit is reached and clamav-daemon closes the stream.
From this moment for each chunk the call to _fwrite()_ will throw an exception and cause a line in the log file.

For the example of a 5GiB video file, 1073741824 exceptions and log file lines will be created.
That needs a lot of time (several minutes), trashes full the log file and is a risc to running out of memory in the filesystem.

Correct behavior would be:
1) Not to use INSTREAM in case of file size over the stream limit
2) In case of exception to stop reading the file and only to log one error in the log file

sidey79:
I've looket into this and found, that there can be a problem with async scans and background scans.

I've not understood all that wrapper things in owncloud, but i see, that the error is caused by fwrite.
The return code of fwrite is not checked, so fwrite is called for every chunk and it seems that it's not checked if fwrite has written all that data to the socket.

I've did a small quick and dirty extension to verify this:

```
public function onAsyncData($data){
                $rc=fwrite(
                                $this->getWriteHandle(),
                                $this->prepareChunk($data)
                );
                if ($rc === false || $rc == 0)
                {
                  $this->initAsyncScan();
                  fwrite(
                                $this->getWriteHandle(),
                                $this->prepareChunk($data)
                  );
                }

        }
```

With these additional lines of code, the rc of fwrite is checked. If data coul'd be written, a new socket will be opened. 

After that, i saw 3 entrys in the log, because fwrite may was able to send some of the 8192 bytes, but not all.
May this can be used as an idea, how to fix the problem or to quickly reduce the amount of logentrys.

DeepDiver1975:
[...]

PR #133, opened 2016-11-02
Add AVMaxFileSize config option

VicDeo:
### Description
- Adds File size limit
- Fixes https://github.com/owncloud/files_antivirus/issues/124 and https://github.com/owncloud/files_antivirus/issues/125

### Changes in Docs:
__Stream Length__ (bytes)  `26214400` (25Mb) by default.  __Should be <=__  than  `StreamMaxLength` ClamAV value
 __File size limit, -1 means no limit__ (bytes)  `-1` by default - prevent background scanner from checking large files (Has no effect while scanning newly uploaded files)

### To test: 
1. set StreamMaxLength in `/etc/clamd.conf` to e.g. 10M 
2. Configure the app in daemon or daemon(socket) mode
3. cat several 100Mb files with EICAR signature
4. Try upload or background scan

### DIFF
--- a/lib/backgroundscanner.php
+++ b/lib/backgroundscanner.php
@@ -24,16 +24,14 @@ class BackgroundScanner {
 	/** @var \OCP\Files\Folder[] */
 	protected $userFolders;
 
-	/**
-	 * @var ScannerFactory
-	 */
+	/** @var ScannerFactory */
 	private $scannerFactory;
 
-	
-	/**
-	 * @var IL10N
-	 */
+	/** @var IL10N */
 	private $l10n;
+
+	/** @var  AppConfig  */
+	private $appConfig;
 
 	/** @var string */
 	protected $currentFilesystemUser;
@@ -46,17 +44,20 @@ class BackgroundScanner {
 	 *
 	 * @param \OCA\Files_Antivirus\ScannerFactory $scannerFactory
 	 * @param IL10N $l10n
+	 * @param AppConfig $appConfig
 	 * @param IRootFolder $rootFolder
 	 * @param IUserSession $userSession
 	 */
 	public function __construct(ScannerFactory $scannerFactory,
 								IL10N $l10n,
+								AppConfig $appConfig,
 								IRootFolder $rootFolder,
 								IUserSession $userSession
 	){
 		$this->rootFolder = $rootFolder;
 		$this->scannerFactory = $scannerFactory;
 		$this->l10n = $l10n;
+		$this->appConfig = $appConfig;
 		$this->userSession = $userSession;
 	}
 	
@@ -69,6 +70,17 @@ class BackgroundScanner {
 		$dirMimeTypeId = \OC::$server->getMimeTypeLoader()->getId('httpd/unix-directory');
 		try {
 			$qb = \OC::$server->getDatabaseConnection()->getQueryBuilder();
+
+			$sizeLimit = intval($this->appConfig);
+			if ( $sizeLimit === -1 ){
+				$sizeLimitExpr = $qb->expr()->neq('fc.size', $qb->expr()->literal('0'));
+			} else {
+				$sizeLimitExpr = $qb->expr()->andX(
+					$qb->expr()->neq('fc.size', $qb->expr()->literal('0')),
+					$qb->expr()->lt('fc.size', $qb->expr()->literal((string) $sizeLimit))
+				);
+			}
+
 			$qb->select(['fc.fileid'])
 				->from('filecache', 'fc')
 				->leftJoin('fc', 'files_antivirus', 'fa', $qb->expr()->eq('fa.fileid', 'fc.fileid'))
@@ -96,9 +108,7 @@ class BackgroundScanner {
 				->andWhere(
 					$qb->expr()->like('fc.path', $qb->expr()->literal('files/%'))
 				)
-				->andWhere(
-					$qb->expr()->neq('fc.size', $qb->expr()->literal('0'))
-				)
+				->andWhere( $sizeLimitExpr )
 			;
 			$result = $qb->execute();
 		} catch(\Exception $e) {
--- a/lib/backgroundscanner.php
+++ b/lib/backgroundscanner.php
@@ -67,50 +67,8 @@ class BackgroundScanner {
 	 */
 	public function run(){
 		// locate files that are not checked yet
-		$dirMimeTypeId = \OC::$server->getMimeTypeLoader()->getId('httpd/unix-directory');
 		try {
-			$qb = \OC::$server->getDatabaseConnection()->getQueryBuilder();
-
-			$sizeLimit = intval($this->appConfig);
-			if ( $sizeLimit === -1 ){
-				$sizeLimitExpr = $qb->expr()->neq('fc.size', $qb->expr()->literal('0'));
-			} else {
-				$sizeLimitExpr = $qb->expr()->andX(
-					$qb->expr()->neq('fc.size', $qb->expr()->literal('0')),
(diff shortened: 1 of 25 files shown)


VicDeo:
@PVince81 please have a look

lib/backgroundscanner.php
```diff
@@ -133,6 +93,68 @@ public function run(){
...
+			->leftJoin('fc', 'files_antivirus', 'fa', $qb->expr()->eq('fa.fileid', 'fc.fileid'))
+			->innerJoin(
+				'fc',
+				'storages',
+				'ss',
+				$qb->expr()->andX(
+					$qb->expr()->eq('fc.storage', 'ss.numeric_id'),
+					$qb->expr()->orX(
+						$qb->expr()->like('ss.id', $qb->expr()->literal('local::%')),
+						$qb->expr()->like('ss.id', $qb->expr()->literal('home::%'))
```
PVince81: unrelated, but reminded me of a potential issue, raised as https://github.com/owncloud/files_antivirus/issues/137

PVince81:
Code looks good.

@owncloud/qa can you help testing ?

@VicDeo Any chance to add unit tests of some sorts ? That quite some complex logic there.
Even a test that only covers the stream stuff using a dummy stream would be nice.

For the docs change, please raise a ticket in the documentation repo.

PVince81:
@VicDeo question: what if the virus signature overlaps with the retry-boundary, can it still be detected properly ?

PVince81:
maybe the unit test could have its own dummy signature and have the mock stream receiver detect it (which simulates whatever clamd does)

VicDeo:
@PVince81 
 > what if the virus signature overlaps with the retry-boundary, can it still be detected properly ?

E.g.  When Stream Length contains 3  chunks. And file has 7.
The scanning will be done as
1st scan request: chunks 1-2-3 
2nd scan request: chunks  3-4-5
3rd scan request: chunks 5-6-7

if a signature is spread across chunks 2-3-4 - the answer is `No`
In general ClamAV doesn't support large files at all. So it is still an improvement :)

Hm. I've just came upon an idea to hide chunk size from UI and make it completely internal.

VicDeo:
@PVince81 I added a dumb avir server but it fails to scan more than one file unfortunately

sidey79:
With Last PR, Chunk Size can be removed from the docs i think.

VicDeo:
@sidey79 thanks, updated

ManfredSteiner:
@VicDeo I test branch fix-125 (b3480aef4067f09756428f09b94a8698b135ecd8).

The *byteCount* should be cleared to zero, if the stream is closed ([scanner.php#L143](https://github.com/owncloud/files_antivirus/blob/b3480aef4067f09756428f09b94a8698b135ecd8/lib/scanner.php#L143)). Test with EICAR Virus Signature file results in *Unsupported Media Type*. File not uploaded.

ms270169:
(Conversation before done in #125)
@VicDeo I have tested branch fix-125 (b3480ae).

The byteCount should be cleared to zero, if the stream is closed ([scanner.php#L143](https://github.com/owncloud/files_antivirus/blob/b3480aef4067f09756428f09b94a8698b135ecd8/lib/scanner.php#L143)). Test with EICAR Virus Signature file results in Unsupported Media Type. File not uploaded.

VicDeo:
@ms270169  byteCount is reset to 0 in  `Scanner::initScanner` that is called just one line below

[scanner.php#L144](https://github.com/owncloud/files_antivirus/blob/b3480aef4067f09756428f09b94a8698b135ecd8/lib/scanner.php#L144)

ms270169:
@VicDeo, you are right (I was wrong), byteCount ist cleared to zero in *initScanner()*. Virus scanning works now as desired.

I tested it with the following configuration:
* testfile: 30000 bytes length, 68 byte EICAR virus signature on position 16360-17027
* config-value *Chunk Size*: 8192 (config not used, because I add the file with plus button)
* used chunk-size: 8192
* config-value *Stream Length*: 20000
* config-value *File size limit*: -1

Virus signature overlaps second and third chunk. Virus is only detected if second chunk is repeated after reinitialisation of stream (if block overlapping is implemented).

**Result:**
Virus detection successful, File not accepted by the cloud-system.

The user gets the info *"Unsupported media Type"*, which is a confusing message. If the other bug, leading to this wrong message could not be fixed (see https://github.com/owncloud/files_antivirus/issues/125#issuecomment-259970947), I would recommend to add a log entry in case of virus detection (for example before [avirwrapper.php#L99](https://github.com/owncloud/files_antivirus/blob/2c3e6ec4407a5f94d201c4bbbe1f31b7c9b92c8a/lib/avirwrapper.php#L99)).

VicDeo:
@ms270169
InvalidContentExeption is mapped tonto  HTTP status 415.  "Unsupported media Type" is a text representation for HTTP status 415.
There is no special HTTP status "User just uploaded a virus" unfortunately.
I can't do anything better on the antivirus side. 403 with message "Forbidden" is neither too much descriptive. :/

VicDeo:
grr. It drive me nuts. All methods of Config mock return `null` on the second test execution.

ms270169:
@VicDeo , maybe you have an idea how to better visualize the message. Because the information about a detected virus is inside the HTTP 415 response. If the message of the response would be shown, everything would be clear for the user...
```
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>OCA\DAV\Connector\Sabre\Exception\UnsupportedMediaType</s:exception>
  <s:message>Virus Eicar-Test-Signature is detected in the file. Upload cannot be completed.</s:message>
</d:error>
```

VicDeo:
I still have no idea why PHPUnit mocks return null for the second testcase. 
I created own mocks instead of mocking with PHPUnit API.
It works this way.

VicDeo:
@PVince81 ready from my POV

PVince81:
@VicDeo from what I see this was tested and only needs a code review, correct ?

VicDeo:
@PVince81 yes
 @ms270169  helped me a lot with a design overview.

PVince81:
The code looks good, great to have tests!

@VicDeo I didn't find where "AVMaxFileSize" is covered / passed in the tests, mind pointing me there ? I guess it probably has another name

VicDeo:
@PVince81 Not covered.
 It is used in background scanner only, because file size is unpredictable while using a stream wrapper.
`BackgroundScanner::getFilesForScan` is the only place in code where is it used.

PVince81:
Ok, fine 👍

VicDeo:
Stable9.1: https://github.com/owncloud/files_antivirus/issues/140

zubanst:
Any back-port plan for 9.0.x? In 9.0.6 I have the same issue described in #125 .

VicDeo:
@zubanst would you like to test https://github.com/owncloud/files_antivirus/pull/144 ?

zubanst:
@VicDeo I would do if I had all the details how to.

SergioBertolinSG:
Updated steps:

```
To reproduce:

1. set StreamMaxLength in /etc/clamd.conf to e.g. 10M
2. Configure the app in daemon or daemon(socket) mode
3. In files_antivirus config in administration change the values:
   Stream Length   to 10000000 bytes
   File size limit, -1 means no limit  to 10000000 bytes
4. cat several 100Mb files with EICAR signature
5. Try upload or background scan
```

Expected: It detects the virus. 
Not expected: It uploads the file.

VicDeo:
@SergioBertolinSG 
Indeed `Stream Length` in files_antivirus configuration should match Clamav  `StreamMaxLength` option.
`File size limit` has no effect on upload.

PR #140, opened 2016-11-22
[Stable9.1] fix 125

VicDeo:
Backport of #133

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ script:
 
   # Run phpunit tests
   - cd tests
+  - php avirserver.php &
   - phpunit --configuration phpunit.xml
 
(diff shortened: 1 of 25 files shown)


PVince81:
@VicDeo 9.0.7 or 9.0.8 ? How risky is this backport ?

@owncloud/qa what do you think ?

VicDeo:
@PVince81 From my POV not too risky, but let's move to the next release. 
Stable9.1 is the same with master, so it will be easier to make backports in future if we keep both in sync.

PVince81:
@davitol please help testing, thanks

davitol:
:+1: WFM

PVince81:
:+1: for the code

PR #141, opened 2016-11-22
Log a warning on uploading an infected file #132

VicDeo:
Ref #132


VicDeo:
@PVince81

PVince81:
Makes sense 👍

PR #143, opened 2016-11-28
Clarify that action is related to background scan only

### LINKED ISSUE #129: "Log only" acts just the same as "Delete"
Hi,

I installed files_antivirus today mostly successfully, however I find that the option "Log only" is not working . With option "Log only", file gets blocked with a message "Virus Eicar-Test-Signature is detected in the file. Upload cannot be completed." so does option "Delete". 

Shouldn't they act differently? I expect that "Log only" gives a warning message, a log info and maybe a confirm button, but not prevent files from uploading.
## My system info is:

Operating System: CentOS 7.1
Web server: Apache 
Database: MySQL
PHP version: 7
ownCloud version: 9.1
Updated from an older ownCloud or fresh install: updated from 8
Antivirus App for files 0.9.0.0 with patch #127 

Thank you very much!

VicDeo:
Closes #129

### DIFF
--- a/templates/settings.php
+++ b/templates/settings.php
@@ -30,11 +30,11 @@ script('files_antivirus', 'settings');
 			<p class="av_max_file_size">
 				<label for="av_max_file_size"><?php p($l->t('File size limit, -1 means no limit'));?></label>
 				<input type="text" id="av_max_file_size" name="avMaxFileSize" value="<?php p($_['avMaxFileSize']); ?>"
-					   title="<?php p($l->t('File size limit in bytes, -1 means no limit'));?>"
+					   title="<?php p($l->t('Background scan file size limit in bytes, -1 means no limit'));?>"
 				/>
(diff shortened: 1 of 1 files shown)


VicDeo:
@PVince81 please proof-read

templates/settings.php
```diff
@@ -34,7 +34,7 @@
 				/>
 				<label for="av_max_file_size" class="a-left"><?php p($l->t('bytes'))?></label>
 			</p>
-			<p class="infected_action"><label for="av_infected_action"><?php p($l->t('Action for infected files found while scanning'));?></label>
+			<p class="infected_action"><label for="av_infected_action"><?php p($l->t('When infected files are found by cron job'));?></label>
```
PVince81: "When infected files were found during a background scan" ?

Or is it too long ?

VicDeo: @PVince81 fits into the setting section

VicDeo:
@PVince81 updated

PVince81:
:+1:

PR #153, opened 2017-03-27
add isScannableSize()

butonic:
This PR prevents scanning files bigger than `appConfig->getAvMaxFileSize`. By default it is -1 (no limit).

The title on the settings page seems to never have been visible. I assume because some tipsy or whatever js replaces the input title with the label title. I aligned it with the label, now.

This changes the behavior, because now if you set a File size limit files bigger than it will not even be sent to the scanner. Arguably, that may very well be what you want when reading the label.

Also may need an udpate of the documentation. cc @settermjd 

- [ ] Tests

### DIFF
--- a/lib/avirwrapper.php
+++ b/lib/avirwrapper.php
@@ -11,7 +11,6 @@ namespace OCA\Files_Antivirus;
 use OC\Files\Filesystem;
 use OC\Files\Storage\Wrapper\Wrapper;
 use \OCP\App;
-use \OCP\IConfig;
 use \OCP\IL10N;
 use \OCP\ILogger;
 use \OCP\Files\InvalidContentException;
@@ -25,7 +24,12 @@ class AvirWrapper extends Wrapper{
 	 * @var array 
 	 */
 	private $writingModes = array('r+', 'w', 'w+', 'a', 'a+', 'x', 'x+', 'c', 'c+');
-	
+
+	/**
+	 * @var AppConfig
+	 */
+	protected $appConfig;
+
 	/**
 	 * @var \OCA\Files_Antivirus\ScannerFactory
 	 */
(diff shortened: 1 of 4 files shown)


butonic:
Ok, the wording for the background scan seems to only have changed in master. in 9.1 it was still `File size limit in bytes, -1 means no limit`

lib/avirwrapper.php
```diff
@@ -136,4 +144,21 @@ private function isWritingMode($mode){
...
+	 * checks the size for webdav PUT requests. defaults to true
+	 * @param $filename
+	 * @return bool
+	 */
+	private function isScannableSize($filename) {
+		$scanSizeLimit = intval($this->appConfig->getAvMaxFileSize());
+		$size = false;
+
+		// PUT via webdav
+		if ($_SERVER['REQUEST_METHOD'] === 'PUT' && isset($_SERVER['CONTENT_LENGTH'])) {
```
VicDeo: @butonic would it be testable this way?

butonic: Yes, the $_SERVER variable can be set during unit testing and private methods can be made public.

DeepDiver1975: Do it. 😆

butonic:
adding sev1 because files cannot be uploaded

michaelstingl:
@butonic @PVince81 Backport to 9.1.x still needed? Or just release with the next (Marketplace) release?

michaelstingl:
@PVince81 It would be cool to have this included in the [next release](https://github.com/owncloud/files_antivirus/milestone/82). I'm not aware of a backport request.

PVince81:
in that case let's not backport. Since this PR is already merged it will be part of the next release of this app in the marketplace

VicDeo:
@PVince81 This is a question that I'd like to have an answer too.
Will we do a marketplace release from a master branch?

PVince81:
@VicDeo master only

PR #154, opened 2017-04-03
Fix markdown

DeepDiver1975:
Markdown formating was a bit broken ....


VicDeo:
:+1: 
@DeepDiver1975 thanks

PR #155, opened 2017-05-15
admin.php is a history

VicDeo:
`admin.php` was removed from the repo so the build process is failing.


PVince81:
@VicDeo please verify if 10.0.1RC3 is affected... From what I see the app is there but I see a Makefile in the folder. So maybe the build worked after all ?

VicDeo:
@PVince81  It seems  makefile from this repo is not used to bundle the app.
RC3 is not affected.
Makefile is sneaked into the release but it's not critical IMO

PVince81:
so does the antivirus app in RC3 work ? (quick smoke test)

VicDeo:
@PVince81 it's possible to enable the app and it detects EICAR on upload

PVince81:
:+1: to merge for 10.0.2 to avoid risks

VicDeo:
Moved to https://github.com/owncloud/files_antivirus/pull/157

PR #168, opened 2017-08-03
[Stable10] Change fileid to bigint

VicDeo:
backport of https://github.com/owncloud/files_antivirus/pull/165
ref https://github.com/owncloud/core/issues/26901


PVince81:
not needed though, as we should release to marketplace only from master

PR #169, opened 2017-08-08
Use migrations

VicDeo:
Go on with https://github.com/owncloud/core/issues/26901

PgSQL dumps 
### Before
```
CREATE TABLE oc_files_antivirus(
  fileid integer NOT NULL,
  check_time integer NOT NULL DEFAULT 0,
  CONSTRAINT oc_files_antivirus_pkey PRIMARY KEY (fileid)
)
WITH (  OIDS=FALSE );
ALTER TABLE oc_files_antivirus  OWNER TO "oc_D";

CREATE TABLE oc_files_avir_status(
  id serial NOT NULL,
  group_id integer NOT NULL DEFAULT 0,
  status_type integer NOT NULL DEFAULT 0,
  result integer NOT NULL DEFAULT 0,
  match character varying(64) DEFAULT NULL::character varying,
  description character varying(64) DEFAULT NULL::character varying,
  status integer NOT NULL DEFAULT 0,
  CONSTRAINT oc_files_avir_status_pkey PRIMARY KEY (id)
)
WITH (  OIDS=FALSE );
ALTER TABLE oc_files_avir_status  OWNER TO "oc_D";
```

### Update
```
CREATE TABLE oc_files_antivirus(
  fileid bigint NOT NULL,
  check_time integer NOT NULL DEFAULT 0,
  CONSTRAINT oc_files_antivirus_pkey PRIMARY KEY (fileid)
)
WITH (  OIDS=FALSE );
ALTER TABLE oc_files_antivirus  OWNER TO "oc_D";


CREATE TABLE oc_files_avir_status(
  id serial NOT NULL,
  group_id integer NOT NULL DEFAULT 0,
  status_type integer NOT NULL DEFAULT 0,
  result integer NOT NULL DEFAULT 0,
  match character varying(64) DEFAULT NULL::character varying,
  description character varying(64) DEFAULT NULL::character varying,
  status integer NOT NULL DEFAULT 0,
  CONSTRAINT oc_files_avir_status_pkey PRIMARY KEY (id)
)
WITH (  OIDS=FALSE );
ALTER TABLE oc_files_avir_status  OWNER TO "oc_D";
```

### Install
```
CREATE TABLE oc_files_antivirus(
  fileid bigint NOT NULL,
  check_time integer NOT NULL DEFAULT 0,
  CONSTRAINT oc_files_antivirus_pkey PRIMARY KEY (fileid)
)
WITH (  OIDS=FALSE );
[...]

PR #170, opened 2017-08-11
Stable10 migrations

VicDeo:
:exclamation:  this PR adds dependency on **ownCloud 10.0.3+**

Backport of #169


PVince81:
@VicDeo please fix failures on travis

PVince81:
restarted build

VicDeo:
@PVince81 
the error is
>App "Antivirus App for files" cannot be installed because the following dep  
  endencies are not fulfilled: ownCloud 10.0.3 or higher is required.

PVince81:
ok, I'll increase the version on master...

appinfo/install.php
```diff
@@ -1,11 +1,4 @@
 <?php
 
-$app = new \OCA\Files_Antivirus\AppInfo\Application();
```
PVince81: assuming this got removed on purpose, obsolete ?

VicDeo: @PVince81 moved into migration.
install script is triggered before migrations, thus here we were trying to insert default rules into the table that did not exist.

PR #175, opened 2017-08-28
Skip chunks - they will be scanned on the final move anyway

VicDeo:
Ref https://github.com/owncloud/enterprise/issues/2227

### DIFF
--- a/lib/avirwrapper.php
+++ b/lib/avirwrapper.php
@@ -67,6 +67,7 @@ class AvirWrapper extends Wrapper{
 		if (is_resource($stream)
 			&& $this->isWritingMode($mode)
 			&& $this->isScannableSize(basename($path))
+			&& strpos($path, 'uploads/') !== 0
 		) {
 			try {
 				$scanner = $this->scannerFactory->getScanner();
--- a/tests/ChunkUploadTest.php
+++ b/tests/ChunkUploadTest.php
@@ -1,7 +1,7 @@
 <?php
 
 /**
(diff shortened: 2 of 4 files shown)


lib/avirwrapper.php
```diff
@@ -67,6 +67,7 @@ public function fopen($path, $mode){
 		if (is_resource($stream)
 			&& $this->isWritingMode($mode)
 			&& $this->isScannableSize(basename($path))
+			&& strpos($path, 'uploads/') !== 0
```
PVince81: can I trick this by having a path "myfolder/something/uploads/infected.exe" ?

PVince81: or is this check very subtle and means: only go there either when "uploads" is not at the start or when "uploads" is not found at all ? A PHP comment would be appreciated to clarify this implicit confusing logic

VicDeo: @PVince81 I think `strpos($path, 'uploads/') !== 0` is self-descriptive as it reads as `path does not start with`

PVince81:
please backport to stable10

VicDeo:
Stable10: https://github.com/owncloud/files_antivirus/pull/176

PVince81:
Oracle setup fail, please investigate separately: https://travis-ci.org/owncloud/files_antivirus/jobs/269767617

VicDeo:
seems to be a hiccup:
>Could not download from "http://pecl.php.net/get/oci8-2.1.7.tgz", cannot download "pecl/oci8" (Connection to `pecl.php.net:80' failed: Connection timed out)

Issue #183, opened 2017-09-01
input fields are not aligned correctly


individual-it:
![image](https://user-images.githubusercontent.com/2425577/29962800-23aed1ce-8f24-11e7-943d-21ab549552b4.png)

VicDeo:
@individual-it browser, screen resolution?
<img width="405" alt="avir" src="https://user-images.githubusercontent.com/991300/30385574-481c3cbe-98b0-11e7-8f96-f04d2d14211c.png">

individual-it:
FF 55 and Chrome 60 both on Debian
Chrome is a little bit different. Also happens on various resolutions
FF:
![ff](https://user-images.githubusercontent.com/2425577/30428952-5dd7aec4-9975-11e7-8ea8-974bf9493d22.gif)
Chrome:
![chrome](https://user-images.githubusercontent.com/2425577/30428948-587c6118-9975-11e7-9d49-8ed78e084e3d.gif)

sorry quality of the gif is not the best

PVince81:
moving to development to work on for next release

individual-it:
seems to be fixed with #187

Issue #178, opened 2017-08-31
missing validation in input fields


individual-it:
the fields 'Host', 'Port', 'Stream Length' & 'File size limit' should be validated better.
All fields seem to accept any character. At least "Port" and "File size limit" do not save strings that do not look like integer, the other fields do.
But you can enter a "minus" in the port field. e.g. "Port : -3310" and it get saved
In 'Host' and 'Stream Length' its possible to enter any kind of rubbish and it gets saved.

PVince81:
moving to development to work on for next release

PVince81:
backlog

individual-it:
This was first fixed (frontend validation) by #187 but it looks like #195 brought that issue back
after #195 I see red marks around invalid fields but still can save the invalid settings 
![image](https://user-images.githubusercontent.com/2425577/34932455-e6d847ac-f9fa-11e7-8edc-f5ce3bafcef0.png)

individual-it:
fixed by #212

PR #187, opened 2017-09-12
Frontend Validation

VicDeo:
Ref #178 : 
>the fields 'Host', 'Port', 'Stream Length' & 'File size limit' should be validated better.
All fields seem to accept any character. At least "Port" and "File size limit" do not save strings that do not look like integer, the other fields do.
But you can enter a "minus" in the port field. e.g. "Port : -3310" and it get saved
In 'Host' and 'Stream Length' its possible to enter any kind of rubbish and it gets saved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+
+### Added
+
+ - Frontend Validation for config fields [187](https://github.com/owncloud/files_antivirus/pull/187)
+
 ## [0.11.1.0] - Unreleased
 
 ### Changed
--- a/css/settings.css
+++ b/css/settings.css
@@ -6,9 +6,13 @@
 
 .section-antivirus label {
 	display: inline-block;
-	min-width: 280px;
+	width: 280px;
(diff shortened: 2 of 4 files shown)


VicDeo:
Regarding hostname: it could be either IP address or domain name including IDNA :\

individual-it:
Error messages show two dots at the end of the sentence
image

individual-it:
otherwise it looks good 👍 
also seems to fix #183

VicDeo:
@individual-it a-ha. it is firefox thing. No extra dot in chrome

VicDeo:
@individual-it  updated

Please note that only master is wired to get translations from Transifex.
If we remove this dot from the messages  for stable branches all translations will be lost.

VicDeo:
... aand one more commit (above) fixing the following issue:

1. Delete port value for daemon mode  
```
php occ config:delete files_antivirus av_port
```
2. try to save settings in Executable mode

expected: settings are saved
actual: invisible port input doesn't pass validation

VicDeo:
@individual-it Somehow I've managed to reproduce #183 by playing with different modes in settings.
The fix is also pushed here.

VicDeo:
@PVince81 backport?

PVince81:
@VicDeo yes

VicDeo:
Stable10: https://github.com/owncloud/files_antivirus/pull/189

Issue #181, opened 2017-09-01
reject file uploads when clamAV is not reachable


individual-it:
When and admin configures the antivirus app, every upload has to be scanned. If the clamAV service is not reachable for whatever reasons the upload cannot be checked and should be rejected.

This was first suggested by @hodyroff in https://github.com/owncloud/core/issues/28870#issuecomment-326413982

PVince81:
I don't think this is a release blocker -> p2

cc @pmaier1 @felixboehm if you think otherwise

PVince81:
the antivirus should likely throw a proper exception to abort the upload

VicDeo:
To reject upload I should to be  100% sure that upload is in progress.
Any kind of `Storage::fwrite` will be rejected If I just throw an exception.

PVince81:
@VicDeo when does the streaming to clamav start ? Even if we started streaming a few bytes and and only then throw an exception it's fine too

Issue #180, opened 2017-09-01
administrator should be notified if clamav not reachable


individual-it:
To see if clamAV is reachable and works the admin has upload a file and check if there any error messages in the log file or he/she has to upload a test virus file e.g the `eicar.com` file with the content `X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*` and see if its rejected.
But it would be handy if the app would check at least the reachability of clamAV when saving the settings

PR #195, opened 2017-10-23
Reject uploads when antivirus is unreachable

VicDeo:
- Reject uploads when antivirus is unreachable/ Fixes https://github.com/owncloud/files_antivirus/issues/181
~Note: it will reject not uploads only but any `fopen` on storage~ - already fixed with #196
-  Adds a connection test after saving the settings. Notify admin  priority label if it is failed. #180 
- Move all scanners into a separate subdir
- Subclass  scan via socket and daemon from parent External
- Cover scanner initialization with unit tests

- [x] TODO:  figure out meaning of `retry` https://github.com/owncloud/core/blob/master/apps/dav/lib/Connector/Sabre/Exception/Forbidden.php#L31
`retry=true` for misconfigured app
`retry=false` for infected files

misconfig-daemon
misconfig-socket
misconfig-executable


VicDeo:
@ownclouders rebase

individual-it:
works like described, but I would like it more if additionally there would be a check when the configuration  gets saved

VicDeo:
@individual-it this is a different issue (#180) but as you wish ;)

VicDeo:
@individual-it  rereview please

Issue #179, opened 2017-08-31
File size limit is not obeyed when using chunking


individual-it:
files of any size get checked regardless of the "File size limit" setting when chunking is used

PVince81:
I don't think this is a release blocker -> p2

cc @pmaier1 @felixboehm if you think otherwise

PVince81:
which chunking ? old webdav chunking or new dav endpoint chunking ? (the latter is with the "uploads/" folder)

Antivirus might need a similar approach like firewall to find out the final file size

individual-it:
chunking on both dav endpoints does not calculate the size correctly

VicDeo:
@individual-it With  which app version?

VicDeo:
<img width="476" alt="limit-v0 10 1 0" src="https://user-images.githubusercontent.com/991300/30037474-cf9f0750-91c3-11e7-8d83-23885aee8ff1.png">

individual-it:
> @individual-it With which app version?

current master branch

PVince81:
moving to development, to look into for the next release @VicDeo

PVince81:
will this need core changes ?

PVince81:
will this need core changes ? can we move forward with this ?

VicDeo:
Fixed with https://github.com/owncloud/files_antivirus/pull/196

individual-it:
reopening as I still have the problem in 0.12.0
reproduce: 
1. set "File size limit" to 100bytes
2. try to upload `eicat_com.zip` (184byte) by chunking e.g.:
```
$ curl --user $user:$pwd http://$host/$path/remote.php/dav/uploads/$user/12345/1 --request PUT --data-binary "@eicar_com.zip_01"
$ curl --user $user:$pwd http://$host/$path/remote.php/dav/uploads/$user/12345/2 --request PUT --data-binary "@eicar_com.zip_02"
$ curl --user $user:$pwd http://$host/$path/remote.php/dav/uploads/$user/12345/3 --request PUT --data-binary "@eicar_com.zip_03"
$ curl --user $user:$pwd http://$host/$path/remote.php/dav/uploads/$user/12345/.file --request MOVE -H "Destination: http://$host/$path/remote.php/dav/files/$user/virus"
```

felixboehm:
@VicDeo please retest, and provide a fix

VicDeo:
I don't see a solution without refactoring the app. 

[...]

PR #196, opened 2017-11-13
Get uploaded file size correctly

VicDeo:
- Properly detect size for chunking/normal upload
- Don't scan chunks for DAV v1/v2
- Ignore calls to `fopen` in case there is no upload (scan file from the storage wrapper only if it is related to the upload)
- Add a debug message to track the decision 

Fixes 
https://github.com/owncloud/files_antivirus/issues/179
https://github.com/owncloud/enterprise/issues/2227

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,4 +44,4 @@ nbproject
 # Mac OS
 .DS_Store
 
-tests/clover.xml
+tests/unit/clover.xml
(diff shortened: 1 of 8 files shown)


VicDeo:
@PVince81 Should work with zsync as zsync implementation still uses `MOVE` method and `OC-Total-Length`  header.
Let's see.

PR #207, opened 2017-12-20
Revise the exception message

settermjd:
This PR updates the exception message to be a bit more readable and refactors it into a constant, as it's used twice in the code.

### DIFF
--- a/lib/AvirWrapper.php
+++ b/lib/AvirWrapper.php
@@ -21,10 +21,12 @@ use Icewind\Streams\CallbackWrapper;
 
 
 class AvirWrapper extends Wrapper{
-	
+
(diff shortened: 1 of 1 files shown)


lib/AvirWrapper.php
```diff
@@ -21,10 +21,12 @@
 
 
 class AvirWrapper extends Wrapper{
-	
+
+	const AV_EXCEPTION_MESSAGE = 'Antivirus app is either misconfigured or inaccessible. %s';
```
VicDeo: @settermjd  Antivirus app is meant to be ownCloud app. It could be misconfigured but can't be inaccessible.
Inaccessible is external antivirus: server/port  or socket. E.g. when clamav service is accidentally stopped.

Issue #208, opened 2018-01-15
chunked files (old algorithm) don't get checked


individual-it:
1. set up the app and make sure it blocks `eicar.com` from http://www.eicar.org/85-0-Download.html
2. setup variables:
```
user=admin
pwd=admin
host=localhost
path=owncloud
```
3. split file for chunking
`split -d -b 50 eicar.com eicar.com_`

4. upload with chunking:
```
curl --user $user:$pwd http://$host/$path/remote.php/webdav/virus-chunking-1234-2-0 --request PUT --data-binary "@eicar.com_00" -H "OC-Chunked:1"
curl --user $user:$pwd http://$host/$path/remote.php/webdav/virus-chunking-1234-2-1 --request PUT --data-binary "@eicar.com_01" -H "OC-Chunked:1"
```

The problem was introduced by #196

VicDeo:
Well, I don't know how to fix it properly.
The only solution is to  scan dav v1 chunks

PVince81:
Didn't it work before for old chunking ? I thought it would check the file during final assembly ?

PVince81:
closed by https://github.com/owncloud/files_antivirus/pull/211

VicDeo:
@PVince81
>Didn't it work before for old chunking ? I thought it would check the file during final assembly ?

yes, since chunks were scanned before. 
But since version 0.12.0 aka "still unreleased" chunks are skipped for performance reasons and only the final file is scanned.

PR #211, opened 2018-01-15
Fix virus detection for public upload and chunking

### LINKED ISSUE #209: public uploads don't get checked
1. set up the app and make sure it blocks `eicar.com` from http://www.eicar.org/85-0-Download.html
2. create an public link with upload permissions
3. upload `eicar.com`

The problem was introduced by #196

VicDeo:
Fixes #209 #208 
Click links above to see reproduction steps


individual-it:
works to block uploads that are uploaded with chunking.
But the max size is not obeyed when using old dav chunking

PVince81:
is max size a separate issue ?

PVince81:
@individual-it please raise separate ticket for max size

individual-it:
reopened #179

PR #213, opened 2018-02-07
skip running background scan when disabled

patrickjahns:
# Motivation

When having user-key encryption enabled, running the background scan is not feasible, since it will just pollute the logs with messages like:

```
{
"reqId":"lzpZJJ71xtqk15b1iXBF",
"level":3,
"time":"2018-02-07T14:16:42+00:00"
,"remoteAddr":"",
"user":"admin",
"app":"files_antivirus",
"method":"--",
"url":"--",
"message":"OCA\\Files_Antivirus\\BackgroundScanner::run, exception: Can not decrypt this file, probably this is a shared file. Please ask the file owner to reshare the file with you."
}
```

This PR allows for disabling background scan by issuing: 
`occ config:app:set files_antivirus av_scan_background --value="false"`

### DIFF
--- a/lib/AppConfig.php
+++ b/lib/AppConfig.php
@@ -20,6 +20,7 @@ use \OCP\IConfig;
 	 * @method string getAvCmdOptions()
 	 * @method string getAvPath()
 	 * @method string getAvInfectedAction()
+	 * @method string getAvScanBackground()
 	 *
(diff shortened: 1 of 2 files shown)


Emil-G:
@patrickjahns , add a config for this to the webinterface?

Issue #177, opened 2017-08-31
'HTTP/1.1 415 Unsupported Media Type' returned when virus found


individual-it:
when a virus is found the antivirus app returns `HTTP 415 UnsupportedMediaType` 
I feel `403 Forbidden` would be more suitable

PVince81:
This might be by design ? I seem to remember @michaelstingl mentioning this in another ticket.

michaelstingl:
I thought https://github.com/owncloud/core/pull/28620 fixed it. (Related: https://github.com/owncloud/enterprise/issues/2141)

PVince81:
https://github.com/owncloud/core/pull/28620 will parse the Sabre exception message as returned by the server. If the message says "Unsupported Media Type" then it will just display that. If we want something else, then the antivirus needs to provide a proper exception message to be displayed.

This ticket is rather about using 403 instead of 415. But I thought 415 was by design already, so we should probably keep it.

michaelstingl:
Oh, I understand. My concerns were more about the error message we receive in the body of the response.

PVince81:
moving to "development" for the next release

PVince81:
@VicDeo please help with the discussion above

PVince81:
moving to backlog

VicDeo:
Fixed with https://github.com/owncloud/files_antivirus/pull/219

PR #219, opened 2018-03-13
A new exception to exclude from logging

sharidas:
The already logs when the file is detected
with virus. So we don't have to log it as
an exception in the server. Hence added a
new exception.

Signed-off-by: Sujith H <sharidasan@owncloud.com>


VicDeo:
Hm. Travis needs a core PR to pass.
Not failing locally, will test manually a bit more.

VicDeo:
Fixes https://github.com/owncloud/files_antivirus/issues/177

VicDeo:
- [ ] Raise min required OC version to 10.0.8 in info.xml to prevent :boom:

PR #220, opened 2018-03-14
Test with stable10

DeepDiver1975:


VicDeo:
@DeepDiver1975 could you please explain it in short?

PR #226, opened 2018-05-17
Fix final size detection for chunked uploads (dav v2 only)

### LINKED ISSUE #179: File size limit is not obeyed when using chunking

VicDeo:
Fixes #179 
Fixes https://github.com/owncloud/enterprise/issues/2145

1. set "File size limit" to 100bytes 
2. upload 3 chunks 40 bytes per chunk
```
user="USER"
pwd="PASSWORD"
host="HOST" # e.g. localhost
path="WEBPATH" # e.g. owncloud
chunk=1234

curl --user $user:$pwd http://$host/$path/remote.php/dav/uploads/$user/$chunk/ --request MKCOL
curl --user $user:$pwd http://$host/$path/remote.php/dav/uploads/$user/$chunk/1 --request PUT --data-binary "@eicar_com.txt_01"
curl --user $user:$pwd http://$host/$path/remote.php/dav/uploads/$user/$chunk/2 --request PUT --data-binary "@eicar_com.txt_02"
curl --user $user:$pwd http://$host/$path/remote.php/dav/uploads/$user/$chunk/3 --request PUT --data-binary "@eicar_com.txt_03"
curl --user $user:$pwd http://$host/$path/remote.php/dav/uploads/$user/$chunk/.file --request MOVE -H "Destination: http://$host/$path/remote.php/dav/files/$user/virus4"
```

### Expected 
`MOVE` request doesn't trigger scanning as 120 bytes are greater than 100
`{"reqId":"ZRjpwP2PYHuAx4Ld9eRi","level":0,"time":"2018-05-21T13:46:58+00:00","remoteAddr":"127.0.0.1","user":"D","app":"files_antivirus","method":"MOVE","url":"\/~deo\/oc-tmp\/remote.php\/dav\/uploads\/D\/117\/.file","message":"File size is 191. av_max_file_size is 100. Scanning is skipped."}
`

### Actual 
it does as file size is considered to be 0:
`{"reqId":"rLxp6i1qRUyv4yYZREhS","level":0,"time":"2018-05-21T14:32:46+00:00","remoteAddr":"127.0.0.1","user":"D","app":"files_antivirus","method":"MOVE","url":"\/~deo\/oc-tmp\/remote.php\/dav\/uploads\/D\/117\/.file","message":"File size is 0. av_max_file_size is 100. Scanning is scheduled."}`

### DIFF
--- a/lib/RequestHelper.php
+++ b/lib/RequestHelper.php
@@ -13,52 +13,111 @@
 
 namespace OCA\Files_Antivirus;
 
+use OC\Cache\CappedMemoryCache;
+use OC\Files\Storage\Storage;
 use \OCP\IRequest;
 
+/**
+ * Used to detect the size of the uploaded file
+ *
+ * @package OCA\Files_Antivirus
+ */
 class RequestHelper {
-
 	/**
 	 * @var  IRequest
 	 */
 	private $request;
 
+	/**
+	 * @var CappedMemoryCache
+	 */
+	private static $fileSizeCache;
+
+	/**
+	 * RequestHelper constructor.
+	 *
+	 * @param IRequest $request
+	 */
 	public function __construct(IRequest $request) {
 		$this->request = $request;
+	}
+
+	/**
+	 * @return CappedMemoryCache
+	 */
+	public function getCache() {
+		if (self::$fileSizeCache === null) {
+			self::$fileSizeCache = new CappedMemoryCache();
+		}
+		return self::$fileSizeCache;
+	}
+
+	/**
+	 * @param string $path
+	 * @param string $size
+	 *
+	 * @return void
+	 */
+	public function setSizeForPath($path, $size) {
+		$this->getCache()->set($path, $size);
 	}
 
 	/**
 	 * Get current upload size
 	 * returns null for chunks and when there is no upload
 	 *
+	 * @param Storage $storage
 	 * @param string $path
 	 *
 	 * @return int|null
 	 */
-	public function getUploadSize($path) {
-		$uploadSize = null;
-
+	public function getUploadSize(Storage $storage, $path) {
 		$requestMethod = $this->request->getMethod();
+
+		// Handle MOVE first
+		// the size is cached by the app dav plugin in this case
+		if ($requestMethod === 'MOVE') {
+			// remove .ocTransferId444531916.part from part files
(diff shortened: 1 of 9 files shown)


lib/RequestHelper.php
```diff
@@ -13,52 +13,106 @@
...
+			$cleanedPath
+		);
+		$cachedSize = $this->getCache()->get($translatedPath);
+		if ($cachedSize > 0) {
+			return $cachedSize;
+		}
 
 		$requestMethod = $this->request->getMethod();
+		// Are we uploading anything?
+		if ($requestMethod !== 'PUT') {
```
PVince81: maybe put this before all the regexp logic ?

lib/RequestHelper.php
```diff
@@ -13,52 +13,106 @@
...
-
-			if ($requestMethod === 'PUT') {
-				$uploadSize = (int)$this->request->getHeader('CONTENT_LENGTH');
-			} else {
-				$uploadSize = (int)$this->request->getHeader('OC_TOTAL_LENGTH');
-			}
-		} else if ($requestMethod === 'PUT' && $isPublicScript) {
-			$uploadSize = (int)$this->request->getHeader('CONTENT_LENGTH');
 		}
+		$uploadSize = (int)$this->request->getHeader('CONTENT_LENGTH');
```
PVince81: are we not using OC_TOTAL_LENGTH for the final MOVE in new chunking ?

VicDeo: @PVince81 it's used but could be omited https://github.com/owncloud/core/blob/master/apps/dav/lib/Upload/ChunkingPlugin.php#L97

lib/RequestHelper.php
VicDeo: mhm. No, but I'd better wrap L75-90 with `if ($requestMethod !== 'MOVE') {`

lib/RequestHelper.php
VicDeo: Well, I answered your question correctly, but neither the question not the answer is applicable here. This part is for non-chunked upload as the new chunking is handled with a Dav plugin above.

PVince81:
@individual-it can you validate this ?

PR #227, opened 2018-05-22
run files_antivirus on drone

patrickjahns:
# Motivation

Bootstrap files_antivirus on drone in order to have acceptance test later on bootstrapped (especially being able to run acceptance tests with objectstore)

### DIFF
--- a/tests/unit/AvirWrapperTest.php
+++ b/tests/unit/AvirWrapperTest.php
@@ -15,6 +15,7 @@ use OCA\Files_Antivirus\Tests\util\DummyClam;
 use Test\Util\User\Dummy;
 
 class AvirWrapperTest extends TestBase {
+
 	const UID = 'testo';
 	const PWD = 'test';
 
--- a/tests/util/DummyClam.php
+++ b/tests/util/DummyClam.php
@@ -65,7 +65,7 @@ class DummyClam {
 
 			$dataChunk = '';
 			do {
-				$dataChunk .= @fread($connection, $chunkSize);
+				$dataChunk .= @fread($connection, $chunkSize - strlen($dataChunk) );
 			} while (is_resource($connection) && strlen($dataChunk) !== $chunkSize);
 
 			$nextBufferSize = strlen($buffer) + strlen($dataChunk);
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,255 @@
+workspace:
+  base: /var/www/owncloud
+  path: apps/files_antivirus
+
+branches: [master, release*, release/*]
+
+pipeline:
+  install-server:
+    image: owncloudci/core
+    pull: true
+    version: ${OC_VERSION}
+    db_type: ${DB_TYPE}
+    db_name: ${DB_NAME}
+    db_host: ${DB_HOST}
+    db_username: ${DB_USERNAME}
+    db_password: ${DB_PASSWORD}
+    db_timeout: 120
+
+  install-app:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+      - cd /var/www/owncloud/apps/files_antivirus
+      - make
+      - cd /var/www/owncloud/
+      - php occ app:list
+      - php occ app:enable files_antivirus
+      - php occ app:list
+      - php occ log:manage --level 0
+
+  code-compliance-check:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+      # currently failing
+      #- make test-php-codecheck
+      - make test-php-lint
+
+  phpunit:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+      # dummy clamav required for unit tests
+      - php /var/www/owncloud/apps/files_antivirus/tests/util/avirserver.php&
+      # wait shortly for the server to be available
+      - wait-for-it 127.0.0.1:5555
+      # required to pass the executable tests
+      - ln -s /var/www/owncloud/apps/files_antivirus/tests/util/avir.sh /usr/bin/clamscan
+      - cd /var/www/owncloud/apps/files_antivirus/tests/unit
+      - phpdbg -d memory_limit=4096M -rr ../../../../lib/composer/bin/phpunit --configuration phpunit.xml
+    when:
+      matrix:
+        TEST_SUITE: phpunit
+
+  codecov:
+    image: plugins/codecov:2
+    secrets: [codecov_token]
+    pull: true
+    files:
+     - 'tests/unit/clover.xml'
+    when:
+      local: false
+      matrix:
+        TEST_SUITE: phpunit
+        PHP_VERSION: 7.0
+
+  scrutinizer:
+    image: owncloudci/php:7.1
+    pull: true
+    commands:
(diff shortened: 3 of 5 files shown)


patrickjahns:
@VicDeo 
Could you give this a look - for master it fails with 
```
There were 2 errors:

1) OCA\Files_Antivirus\Tests\unit\AvirWrapperTest::testInfectedFwrite
fclose() expects parameter 1 to be resource, object given

/var/www/owncloud/lib/private/legacy/util.php:437
/var/www/owncloud/lib/private/legacy/util.php:429
/var/www/owncloud/lib/private/legacy/util.php:406
/var/www/owncloud/lib/private/User/Session.php:430
/var/www/owncloud/lib/private/User/Session.php:504
/var/www/owncloud/lib/public/Events/EventEmitterTrait.php:50
/var/www/owncloud/lib/private/User/Session.php:516
/var/www/owncloud/lib/private/User/Session.php:316
/var/www/owncloud/apps/files_antivirus/tests/unit/AvirWrapperTest.php:54

2) OCA\Files_Antivirus\Tests\unit\ItemTest::testRead
fclose() expects parameter 1 to be resource, object given

/var/www/owncloud/lib/private/legacy/util.php:437
/var/www/owncloud/lib/private/legacy/util.php:429
/var/www/owncloud/lib/private/legacy/util.php:406
/var/www/owncloud/lib/private/User/Session.php:430
/var/www/owncloud/lib/private/User/Session.php:504
/var/www/owncloud/lib/public/Events/EventEmitterTrait.php:50
/var/www/owncloud/lib/private/User/Session.php:516
/var/www/owncloud/lib/private/User/Session.php:316
/var/www/owncloud/apps/files_antivirus/tests/unit/ItemTest.php:36

ERRORS!
Tests: 26, Assertions: 43, Errors: 2.
```

I've seen this error before and it seems related to https://github.com/owncloud/core/pull/31285#issuecomment-390181706

VicDeo:
@patrickjahns these tests need DummyClam running in background https://github.com/owncloud/files_antivirus/blob/master/tests/util/avirserver.php

patrickjahns:
@VicDeo 
it is running - see: https://github.com/owncloud/files_antivirus/pull/227/files#diff-3216dfff0ed3e301453e6799e8c367e2R44

patrickjahns:
For some odd reason this stuff seems like 🔮 - sometimes it works, sometimes it doesn't and I can't reproduce locally (as it always works there - no matter if docker containers or not) - makes me wonder if again our dns servers are making me go crazy

tests/util/DummyClam.php
```diff
@@ -71,7 +71,7 @@ protected function handleConnection($connection){
 				echo 'reading dataChunk' . PHP_EOL;
 				$dataChunk .= @fread($connection, $chunkSize);
 				echo 'dataChunk size '.strlen($dataChunk). PHP_EOL;
-			} while (is_resource($connection) && strlen($dataChunk) < $chunkSize);
+			} while (is_resource($connection) && strlen($dataChunk) !== $chunkSize);
```
VicDeo: I should say that chunkSize is a packed binary that is sent before the chunk itself. so it is not that correct techically.

VicDeo: this is what clamav expects as well

patrickjahns: @VicDeo 
tbh - I was just pocking your implementation to find the place where it fails - and at first it seemed like this because of:

image

Unfortunately for whatever reason - I can't access the drone logs - and that loop fills up the log size of drone.

I've also seen that the chunksize is unpacked wrongly

patrickjahns:
@VicDeo 

Unfortunately the dummy test part is not reliable on drone - need to find a different way.
Using clamav+eicar directly in the tests results in:

```
[phpunit:L20:13s] 1) OCA\Files_Antivirus\Tests\unit\AvirWrapperTest::testInfectedFwrite
[phpunit:L21:13s] Failed asserting that exception of type "\OCP\Files\FileContentNotAllowedException" is thrown.
[phpunit:L22:13s] 
[phpunit:L23:13s] 2) OCA\Files_Antivirus\Tests\unit\AvirWrapperTest::testBigInfectedFwrite
[phpunit:L24:13s] Failed asserting that exception of type "\OCP\Files\FileContentNotAllowedException" is thrown.
[phpunit:L25:13s] 
[phpunit:L26:13s] 3) OCA\Files_Antivirus\Tests\unit\AvirWrapperTest::testInfectedFilePutContents
[phpunit:L27:13s] Failed asserting that exception of type "\OCP\Files\ForbiddenException" is thrown.
[phpunit:L28:13s] 
```

Which I guess is to be expected.

Other options: 
- skip the tests on CI (less favorable) 
- ???

tests/util/DummyClam.php
```diff
@@ -50,26 +50,36 @@ protected function handleConnection($connection){
...
 			$dataChunk = '';
+			$limit = 10;
 			do {
+				echo 'reading dataChunk' . PHP_EOL;
 				$dataChunk .= @fread($connection, $chunkSize);
-			} while (is_resource($connection) && strlen($dataChunk) !== $chunkSize);
+				echo 'dataChunk size '.strlen($dataChunk). PHP_EOL;
+				$limit--;
+			} while (is_resource($connection) && strlen($dataChunk) !== $chunkSize && $limit > 0);
+			if ($limit === 0) break;
```
patrickjahns: @VicDeo 
this very dirt hack of limiting will make the PR green - but it was more intended to abort termination to not fill drone logs

patrickjahns: well but it breaks for a different test then:

```
chunksize = 8192
reading dataChunk
dataChunk size 8188
reading dataChunk
dataChunk size 16380
reading dataChunk
dataChunk size 16422
reading dataChunk
dataChunk size 16422
reading dataChunk
dataChunk size 16422
reading dataChunk
dataChunk size 16422
reading dataChunk
dataChunk size 16422
reading dataChunk
dataChunk size 16422
reading dataChunk
dataChunk size 16422
reading dataChunk
dataChunk size 16422
16422 524288
warning - limit reached during fread - using abortScanned OK
```

tests/util/DummyClam.php
```diff
@@ -65,7 +65,7 @@ protected function handleConnection($connection){
 
 			$dataChunk = '';
 			do {
-				$dataChunk .= @fread($connection, $chunkSize);
+				$dataChunk .= @fread($connection, $chunkSize - strlen($dataChunk) );
```
patrickjahns: after all the debugging - this was the thing to be taken care of. Needed some thinking of what is going on

tests/unit/AvirWrapperTest.php
```diff
@@ -15,6 +15,7 @@
 use Test\Util\User\Dummy;
 
 class AvirWrapperTest extends TestBase {
+
```
VicDeo: @patrickjahns violates our phpcs style rules ;)

patrickjahns: Will add code-style in a seperate PR :-)

VicDeo: @patrickjahns there is no need to add an empty line here anyway. As there are no other changes in this file.

Issue #224, opened 2018-05-09
php-cs-fixer


mmattel:
@DeepDiver1975 
Any plans to implement ``php-cs-fixer`` as we have in core ?

VicDeo:
done

PR #231, opened 2018-05-24
Add coding standard

patrickjahns:
Addresses #224

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -35,6 +35,7 @@ pipeline:
       # currently failing
       #- make test-php-codecheck
       - make test-php-lint
+      - make test-php-style
 
   phpunit:
     image: owncloudci/php:${PHP_VERSION}
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /build
+/vendor
 
 # just sane ignores
 .*.sw[po]
@@ -45,3 +46,4 @@ nbproject
 .DS_Store
 
 tests/unit/clover.xml
+.php_cs.cache
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,0 +1,11 @@
+<?php
+
+$config = new OC\CodingStandard\Config();
+
+$config
+    ->setUsingCache(true)
+    ->getFinder()
+    ->exclude(['vendor','build'])
+    ->in(__DIR__);
+
+return $config;
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,19 @@
 # Makefile for building the project
 
 app_name=files_antivirus
+build_dir=$(CURDIR)/build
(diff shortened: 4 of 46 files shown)


patrickjahns:
Interesting part - it touched translations - how is this usually handled @DeepDiver1975 @VicDeo - do we need to ignore them?

patrickjahns:
@VicDeo 
For the dev-dependencies? Yup - need to fix the platform first though

VicDeo:
https://drone.owncloud.com/owncloud/files_antivirus/92/18

```
Problem 1
    - Installation request for doctrine/annotations v1.6.0 -> satisfiable by doctrine/annotations[v1.6.0].
    - doctrine/annotations v1.6.0 requires php ^7.1 -> your PHP version (5.6.35) does not satisfy that requirement.
```
and so on...

VicDeo:
php files in `l10` look like leftovers as we moved to js a while ago. @DeepDiver1975 correct me if I'm wrong

mmattel:
why do we post allow php 5.6 if core 10.1 will anyhow need +7.1 ?
would´nt it be better to for files_antivirus (and for other related repros) to use 7.1 consistently instead?

VicDeo:
@mmattel I wouldn't mind if someone take over maintaining a separate OC 10 compatible version. 
As for now the app should be cross-compatible with master/stable10.

mmattel:
This gives us the timeline of approx 6 months up to 10.1 will be available.
If we do not make this repro compatible to PHP 7.1 in this time frame, we have to disable it, independent of who maintains it ...
As far I know core master / stable10 requires already PHP 7.1

VicDeo:
@mmattel The coding standard itself (as a set of rules) is not affected, it's all about the library that enforces these rules. It has a dozen of dependencies and we just kindly ask the composer to use versions that don't fail our builds for PHP5.6

patrickjahns:
@mmattel 

The explanation by @VicDeo is quite straight to the point. Pining the php platform version to 5.6 allows us to tell composer: please ensure that all libraries are compatible with php5.6. The coding-standard is not affected by this change in any way - as we currently do not enforce the usage of any language specifics only available in php7.1+ 

As ownCloud 10 currently is compatible with php5.6 - 7.1 (mind 7.2 is still "preview") - we need to find the most common viable libraries. 

As `files_antivirus` doesn't have any composer dependencies itself - nothing is affected by that platform version

mmattel:
@patrickjahns thanks for the detailed view. Would this explanation help others when documented in the coding-standard readme.md?

patrickjahns:
@mmattel 
Not seen this in any other project - not sure what you want to achieve here

@VicDeo @phil-davis 
approve or still changes needed?

VicDeo:
Let's go

Issue #228, opened 2018-05-23
API (acceptance) tests 


patrickjahns:
It would be great to bootstrap api acceptance tests for files_antivirus

### Prerequisites
- [x] bootstrap drone testing + enable clamav in drone @patrickjahns
 
### Scenarios
- [x] Uploading via regular upload #232 
- [x] Uploading via old chunking #232 
- [x] Uploading via new chunking endpoint #232 
- [x] Public uploads (issue #209 ) (PR #244)
- [x] File-size limit when chunking (issue #179 ) (PR #245)

When we have basic tests implemented, this serves additionally as a basis for automated testing `files_antivirus` combined with objectstorage

patrickjahns:
For drone - we can for now use `dinkel/clamavd` as clamav container. If needed we can fork and build it in https://github.com/owncloud-ci

individual-it:
some mote tests in #246
closing this now?

PR #232, opened 2018-05-28
API acceptance test basic infrastructure

phil-davis:
Just the basic folder structure and minimal files needed for acceptance tests.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -36,6 +36,9 @@ pipeline:
       #- make test-php-codecheck
       - make test-php-lint
       - make test-php-style
+    when:
+      matrix:
+        TEST_SUITE: phpunit
 
   phpunit:
     image: owncloudci/php:${PHP_VERSION}
@@ -74,7 +77,34 @@ pipeline:
     when:
       local: false
       matrix:
+        TEST_SUITE: phpunit
         PHP_VERSION: 7.0
+
+  fix-permissions:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+      - cd /var/www/owncloud/
+      - chown www-data * -R
+    when:
+      matrix:
+        TEST_SUITE: api-acceptance
+
+  api-acceptance-tests:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+      # dummy clamav required for unit tests
+      - php /var/www/owncloud/apps/files_antivirus/tests/util/avirserver.php&
+      # wait shortly for the server to be available
+      - wait-for-it 127.0.0.1:5555
+      # required to pass the acceptance tests which used the "dummy" clamscan
+      - ln -s /var/www/owncloud/apps/files_antivirus/tests/util/avir.sh /usr/bin/clamscan
+      - cd /var/www/owncloud/apps/files_antivirus/tests/acceptance
+      - su-exec www-data ./run.sh
+    when:
+      matrix:
+        TEST_SUITE: api-acceptance
 
   owncloud-log:
     image: owncloudci/php:${PHP_VERSION}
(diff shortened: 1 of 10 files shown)


patrickjahns:
additional cases / scenarios to consider:

old-style chunking: https://github.com/owncloud/files_antivirus/issues/208 
public uploads: https://github.com/owncloud/files_antivirus/issues/209 
file-size limit when chunking https://github.com/owncloud/files_antivirus/issues/179

tests/acceptance/features/apiAntivirus/antivirusMain.feature
```diff
@@ -0,0 +1,101 @@
...
+		And user "user0" uploads new chunk file "2" with "is a nasty " to id "chunking-42" using the API
+		And user "user0" uploads new chunk file "3" with "virus" to id "chunking-42" using the API
+		And user "user0" moves new chunk file with id "chunking-42" to "/myChunkedFile.txt" with size 28 using the API
+		Then as "user0" the file "/myChunkedFile.txt" should not exist
+		And the last lines of the log file should contain log-entries containing these attributes:
+			| user  | app             | method | message               |
+			| user0 | files_antivirus | MOVE   | Infected file deleted |
+
+	Scenario: A small file without a virus can be uploaded in chunks using API version 2 and new DAV path
+		Given using API version "1"
```
patrickjahns: shouldn't this be 2 - as the scenario state using API version 2 and new DAV path

phil-davis: API v1 supports "new DAV", and so does API v2. So I tested them both (like was done in core for some tests), just to be sure. See next scenario.

phil-davis:
This currently needs some bonus acceptance test step code that is coming in core.
https://github.com/owncloud/core/pull/31614 and its backport need approval and merge.
Then that, plus https://github.com/owncloud/core/pull/31608 and its backport (both already merged in core), need to get in the "nightly" image.

After that, CI should pass.

.drone.yml
```diff
@@ -37,16 +37,21 @@ pipeline:
...
+  setup-antivirus-environment:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
     commands:
       # dummy clamav required for unit tests
       - php /var/www/owncloud/apps/files_antivirus/tests/util/avirserver.php&
       # wait shortly for the server to be available
       - wait-for-it 127.0.0.1:5555
-      # required to pass the executable tests
+      # required to pass the executable tests and acceptance tests
```
patrickjahns: this needs to be done in the container where the tests run to properly work. 
Otherwise you do in 
- container A - a symlink at `/usr/bin/clamscan` 
- containB doesn't know about `/usr/bin/clamscan` 

only drone workspace is shared between containers

patrickjahns: additional - changing the container for the remote server, also means the unit tests need to be injected with the host name of the container where the dummy clamav server is running

phil-davis: Thanks for the clues!

phil-davis:
This seems good as a first set of acceptance test infrastructure with basic upload with/without chunking tests. @patrickjahns if it seems OK, then I suggest merge it and then I will add more tests in a next PR.

tests/acceptance/features/apiAntivirus/antivirusMain.feature
```diff
@@ -0,0 +1,101 @@
...
+		And user "user0" uploads new chunk file "2" with "is a nasty " to id "chunking-42" using the API
+		And user "user0" uploads new chunk file "3" with "virus" to id "chunking-42" using the API
+		And user "user0" moves new chunk file with id "chunking-42" to "/myChunkedFile.txt" with size 28 using the API
+		Then the last lines of the log file should contain log-entries containing these attributes:
+			| user  | app             | method | message               |
+			| user0 | files_antivirus | MOVE   | Infected file deleted |
+		And as "user0" the file "/myChunkedFile.txt" should not exist
+
+	Scenario: A small file without a virus can be uploaded in chunks using API version 2 and new DAV path
+		Given using API version "1"
```
patrickjahns: sorry to point again - what is the reason for stating `version 2` in the scenario description
> A small file without a virus can be uploaded in chunks using API version 2 and new DAV path

But then using v1 in Given step
> Given using API version "1"

patrickjahns:
I wonder if it is really necessary to change the skeleton directory for testing the app?

Also thinking further - the `run.sh` parts here are again a large copy+paste from the parts in core. 
we might think for the future, to move those general method into a "library" shell script that can be sourced for the apps to have all the bootstrap steps available ? what do you think?

phil-davis:
Yes, we certainly need to sort out removing all the duplicated ``run.sh`` "almost copies" form all the apps. I would like to do that after I get the "all-encompassing" ``run.sh`` merged in ``core`` that does both API and UI tests. Then we can have a single pass through the apps to "fix all the things".

phil-davis:
The skeleton directory code in ``run.sh`` just came "for free" with the other generic stuff when I copied from another app to get started. And when we use a common ``run.sh`` from core, then all that stuff will happen by default anyway. So I suggest we just leave it here for now.

patrickjahns:
Once merged - please update https://github.com/owncloud/files_antivirus/issues/228 accordingly

tests/acceptance/features/apiAntivirus/antivirusMain.feature
phil-davis: Well spotted - I put that cut-paste-forget-to-edit error there just to see if reviewers were reviewing well ;)
Fixed, thanks

PR #233, opened 2018-06-05
test on objectstore

patrickjahns:
# Motivation

new objectstore app release approach - this app has storage relevant parts and should be tested on objectstore


patrickjahns:
failure on stable10 is to be expected

patrickjahns:
I've rebased to not yet test on stable10 - should now be green, and once we move forward on stable10n we enable it for stable10

PR #234, opened 2018-06-08
Remove deprecated SnippetAcceptingContext

phil-davis:
like in core https://github.com/owncloud/core/pull/31687


phil-davis:
It is a mindless change - I will merge without approval, since I can :)

PR #244, opened 2018-07-20
acceptance tests for public uploads

individual-it:
see #228


individual-it:
tests failing because we have to wait for PRs to be merged into core and the qa builds to be build tomorrow

individual-it:
skipped one test that fails in objectstore because of https://github.com/owncloud/files_primary_s3/issues/69

PR #245, opened 2018-07-20
test to check the file size setting

individual-it:
check if the file size limit is obeyed see #228

tests failing because we have to wait for PRs to be merged into core and the qa builds to be build again


individual-it:
@phil-davis fixed the typo

PR #251, opened 2018-08-23
Install and enable testing app

phil-davis:
Just to prove that CI works, while the QA tarball+testing app issues are worked on.


phil-davis:
Working like this. QA tarballs are now in the process of being fixed/rebuilt.

PR #253, opened 2018-08-27
fixup modifyAppConfigs method name

phil-davis:
to match the change in core https://github.com/owncloud/core/pull/32452


phil-davis:
CI will need to be re-run when the core change gets into the next QA tarball build.

PR #254, opened 2018-08-27
su-exec www-data is not needed for acceptance tests

phil-davis:


phil-davis:
```
Could not set mail_domain on http://localhost:8080/ocs/v2.php/apps/testing/api/v1/occ.
```
happening on a few repos. Needs looking into, will be some common issue to fix.

phil-davis:
PR #257 helped a lot here. It runs the server as a proper server, and the acceptance tests run as their own thing separate from the server.
Previously it was running a local PHP dev server in-line with the acceptance test ``.run.sh`` - and that meant that ``www-data`` user was required so that the PHP dev server could see the codde to serve.

PR #257, opened 2018-08-29
Use server for acceptance tests

phil-davis:
I noticed that the acceptance tests here have the owncloud log mixed through the middle of the output, and are using the PHP dev server.

Make them like other app repos (e.g. I used drone.yml examples from notifications and ransomware).

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -44,7 +44,15 @@ pipeline:
       - php occ app:list
       - php occ app:enable files_antivirus
       - php occ app:list
+      - php occ config:system:set trusted_domains 1 --value=server
       - php occ log:manage --level 0
+
+  owncloud-log:
+    image: owncloud/ubuntu:16.04
+    detach: true
+    pull: true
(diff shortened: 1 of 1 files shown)


phil-davis:
I noticed this when looking at the ``su-exec www-data`` stuff.
It will be easier to do this PR first, then see what to do next.

.drone.yml
```diff
@@ -44,8 +44,16 @@ pipeline:
       - php occ app:list
       - php occ app:enable files_antivirus
       - php occ app:list
+      - php occ config:system:set trusted_domains 2 --value=server
```
individual-it: why `2`?

phil-davis: Cut-paste from elsewhere. I think 1 is "localhost" by default, and I am not even sure if this is zero-based. I  will change it to 1.

phil-davis:
``trusted_domains`` index changed to 1.
@individual-it please review again.

PR #259, opened 2018-09-10
Disable background scans by default

micbar:
## Current behavior
1) Enable Antivirus App
2) Background Scanning on the file System is enabled.
3) Scan mode is "Executable"
This results in high CPU load on the web server

## Expected behavior
1) Background Scans should not be enabled by default


VicDeo:
Using  clamav in executable mode is discouraged in general https://doc.owncloud.org/server/10.0/admin_manual/configuration/server/antivirus_configuration.html#executable

Besides high CPU load for av cron job it slows down uploading significally. 
And that part could not be solved.

PVince81:
@micbar @cdamken does the last comment invalidate the issue or should we still move forward ?

PVince81:
closing as this is not recommended

micbar:
@PVince81 
As I understand it, the default config is not recommended and needs to be changed.

micbar:
@VicDeo what would you recommend?

PVince81:
@micbar no, what is not recommended is the usage of "executable mode". one needs to setup "network mode" instead. so it felt like this change of default becomes obsolete / not needed if you switch to "network mode"

VicDeo:
@micbar default config is slow but always working. It could be switched into the **socket** mode but clamd won't be configured to run via socket. The same applies to the **daemon** mode.
All uploads are blocked if the app is misconfigured ans this should also be taken into consideration. 
So there are two options: working and suboptimal defaults or optimized but non-working defaults.

micbar:
Ok, we had a customer case, where the default setting made the instance unusable, but if there is no other "working" default, we obviuosly have not other option.

PR #260, opened 2018-09-17
Add a diagnosis message to bgjob

VicDeo:
Adding an info logtext that a user/fileid/path gets scanned.

### DIFF
--- a/lib/BackgroundScanner.php
+++ b/lib/BackgroundScanner.php
@@ -189,6 +189,10 @@ class BackgroundScanner {
 		$this->initFilesystemForUser($owner);
 		$view = Filesystem::getView();
 		$path = $view->getPath($fileId);
+		\OC::$server->getLogger()->debug(
+			"About to scan file of user {$owner} with id {$fileId} and path {$path}",
+			['app' => 'files_antivirus']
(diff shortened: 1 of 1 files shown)


VicDeo:
@cdamken I think should be enough

lib/BackgroundScanner.php
```diff
@@ -189,6 +189,10 @@ protected function scanOneFile($owner, $fileId) {
 		$this->initFilesystemForUser($owner);
 		$view = Filesystem::getView();
 		$path = $view->getPath($fileId);
+		\OC::$server->getLogger()->info(
```
PVince81: not debug ?

lib/BackgroundScanner.php
```diff
@@ -189,6 +189,10 @@ protected function scanOneFile($owner, $fileId) {
...
+			"About to scan file of a user {$owner} with id {$fileId} and path {$path}",
```
mmattel: @VicDeo if you do a change I propose a slight text adoption:
`About to scan file of a user` -->  `About to scan file of user`
(No a)
And if you have the filename it helpful for the admin.

Issue #272, opened 2018-11-22
Move phpunit.xml to root folder


phil-davis:
Check these things:

- [x] move unit tests into ``test/unit`` folder
- [x] put ``phpunit.xml`` into the root folder
- [x] review ``Makefile`` etc to make the test infrastructure "standard"
- [x] anything else to make the CI and drone "standard"

Part of https://github.com/owncloud/QA/issues/582

PR #273, opened 2018-11-22
Move phpunit.xml to root folder of repo

phil-davis:
Issue #272 

- [x] Move phpunit.xml to root folder of repo
- [x] add PHP unit targets to `Makefile`
- [x] use those targets in drone
- [x] implement coding-standard checks


phil-davis:
I made this a lot like the ``Makefile`` and CI of ``password_policy``

Please review and see what else I should be adjusting.

composer.json
```diff
@@ -6,7 +6,15 @@
             "php": "5.6"
         }
     },
+    "require": {
+    },
     "require-dev": {
-        "owncloud/coding-standard": "^1.0"
+        "phpunit/phpunit": "^5.7",
```
DeepDiver1975: why? core's phpunit is used - isn't it?

phil-davis: I pasted this from ``password_policy``
I will try without it.

phil-davis: phpunit works without this :+1:

phil-davis:
@DeepDiver1975 @patrickjahns @PVince81 this is ready for review again.
I got rid of the local ``phpunit`` and it uses ``phpunit`` from core, in the same way that ``data_exporter`` is doing.

.gitignore
```diff
@@ -1,5 +1,6 @@
 /build
 /vendor
+vendor-bin/**/composer.lock
```
patrickjahns: missing `vendor-bin/**/vendor` - otherwise that part will get accidentally committed

phil-davis: fixed
I also fixed it in password_policy. It is so difficult to get one app as "gold" that can be copied from!

phil-davis:
I pasted in the "standard" rules for "all" the make targets. Then they are there "ready to go".
@patrickjahns do you want me to remove them?

phil-davis:
I added coding standard matrix entry for PHP 5.6. This verifies that the code does not have any PHP7-specific stuff in it.

phil-davis:
Rebased to trigger drone again.

phil-davis:
@patrickjahns @DeepDiver1975 drone is passing.
Ready for review again.

.drone.yml
```diff
@@ -45,91 +48,99 @@ pipeline:
       - php occ app:enable files_antivirus
       - php occ app:enable testing
       - php occ app:list
-      - php occ config:system:set trusted_domains 1 --value=server
+      - php occ config:system:set trusted_domains 1 --value=owncloud
```
patrickjahns: Any particular reason for the change in container naming?

.drone.yml
```diff
@@ -45,91 +48,99 @@ pipeline:
...
+      matrix:
+        NEED_SERVER: true
 
-  code-compliance-check:
+  owncloud-coding-standard:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
     commands:
       # currently failing
       #- make test-php-codecheck
```
patrickjahns: We can drop that one I suppose - as `test-php-codecheck` is using ownclouds code checker which should rather be deprecated and implemented in a tool

.drone.yml
```diff
@@ -45,91 +48,99 @@ pipeline:
...
       matrix:
-        TEST_SUITE: phpunit
+        TEST_SUITE: owncloud-coding-standard
 
-  phpunit:
+  phpunit-tests:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
+    environment:
+      - PHP_VERSION=${PHP_VERSION}
```
patrickjahns: What is the reason for `PHP_VERSION` ? Where do we use this variable and why is it important?

.drone.yml
phil-davis: just to do it the same as many of the other app repos

.drone.yml
```diff
@@ -144,8 +155,8 @@ services:
   oci:
     image: deepdiver/docker-oracle-xe-11g
```
patrickjahns: @DeepDiver1975 - did we move the oracle image to `owncloud-ci` yet?

.drone.yml
phil-davis: It is not important. I will remove it and make a note to myself to remove it everywhere else.

It was copied from ``.drone.yml`` in password_policy, files_texteditor and probably a few other apps.
This is where we need a "gold" standard to copy-paste from. Because we keep having to think over-and-over if the code is optimal.

.drone.yml
phil-davis: If it is the same in every repo, then it is so easy to cut-paste blocks of code into ``.drone.yml`` files and minimize the amount of thinking needed about what little things need to be adjust and where.

Of course, we can decide that ``server`` is going to be the standard name for the ownCloud server service, and adjust the apps that currently call it ``owncloud``

vendor-bin/owncloud-codestyle/composer.json
```diff
@@ -0,0 +1,10 @@
+{
+  "config" : {
+    "platform": {
+      "php": "5.6.37"
```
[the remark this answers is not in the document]
phil-davis: I wonder why this has been done and copy-pasted like this to lots of app repos?

For "real" dependencies that are used by the actual app (or core stable10) we have been mentioning PHP 5.6 so that whenever someone does ``composer update``, even though they may be currently running PHP  7.*, composer will try to find a combination of dependencies that will support PHP 5.6

In this case we do not really care - ``composer`` can find the necessary dependencies to go with whatever PHP  version we happen to be running at the time we do code style check.

.drone.yml
patrickjahns: Seems fine to have the same naming everywhere - whichever (owncloud vs server) is fine for me. 

I am a bit concerned about the copy/cut + paste part - hopefully we can agree that change to CI should be done by trying to understand how it works. Otherwise things will get copy+pasted around without people knowing what actually is going on -> false sense of security

vendor-bin/owncloud-codestyle/composer.json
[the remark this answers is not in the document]
patrickjahns: Please point me towards the initial starting point - the platform limitation here shouldn't have happened. 

Maybe someone had issues locally when developing on 7.x and then switching to 5.6 ?

patrickjahns:
Thinking about https://github.com/owncloud/files_antivirus/pull/273/files/e6210e7ee74237f4e7b4ce77f2f49f2f513ae737#diff-4284501ea35da71f86e08fa8e56fdead 
We probably require a rule that cleans up dependencies ( remove vendor / - vendor-bin/*/vendor - vendor-bin/*/composer.lock) folders to avoid platforming pinning when not necessary.

Opinions?

.drone.yml
phil-davis: I agree, it is good that people understand what they are doing!

But also, if they do it a bit different in each repo then we end up with long reviews, as the reviewers need to ask why. If we make things the same, then [hopefully|maybe some day] review of this sort of thing will be simple.

(and course if there was a way for an app to "suck in" lots of this "standard" CI script infrastructure, that would be even better - no cut-pasting of blocks of script)

vendor-bin/owncloud-codestyle/composer.json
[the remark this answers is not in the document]
phil-davis: That will be difficult to pinpoint. But I notice that ``data_exporter`` has the very simple:
```
{
    "require": {
        "owncloud/coding-standard": "^1.0"
    }
}
```
So I guess pinning the PHP version snuck in somewhere.

phil-davis:
> We probably require a rule that cleans up dependencies ( remove vendor / - vendor-bin//vendor - vendor-bin//composer.lock) folders to avoid platforming pinning when not necessary.

Yes, that would be useful - specially when dropping back to PHP5.6 in a local dev environment to investigate weird stuff. I raised issue https://github.com/owncloud/QA/issues/604 because this is a cross-app thing. We need to decide the ``clean*`` target names and what they do once, then implement in all relevant repos.

vendor-bin/owncloud-codestyle/composer.json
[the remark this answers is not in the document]
phil-davis: I removed the ``config`` ``platform`` section. CI still passes.

phil-davis:
@patrickjahns any more code changes needed for this PR?

.drone.yml
```diff
@@ -45,92 +48,97 @@ pipeline:
...
     when:
       matrix:
-        TEST_SUITE: api-acceptance
+        TEST_SUITE: owncloud-coding-standard
 
-  fix-permissions:
+  phpunit-tests:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
+    environment:
```
phil-davis: @patrickjahns yesterday we removed:
```
    - PHP_VERSION=${PHP_VERSION}
```
@individual-it has commented at https://github.com/owncloud/activity/pull/660#discussion_r237371096

--------
it does make a difference if you run drone locally by `drone exec` and specify the PHP version e.g.
`PHP_VERSION=7.2 drone exec` without that line it would not pick the correct version
--------

thoughts?

patrickjahns: Seems to be strange, as the image `image: owncloudci/php:${PHP_VERSION}` specifies the container to use. 

The `environment` key is only to inject the variable in the environment inside the container. Where does that environment need to be set inside the container?

phil-davis: @individual-it your input is needed - please explain

individual-it: I was wrong, its not needed here.
I had problems running core acceptance tests
running `PHP_VERSION=7.1 TEST_SUITE=api BEHAT_SUITE=apiCapabilities DB_TYPE=mariadb USE_SERVER=true INSTALL_SERVER=true CHOWN_SERVER=true drone exec` in core
gives me:
`Running all API tests tagged ~@skipWhenTestingRemoteSystems&&~@skipOnOcV11&&~@skipOnOcV11.0&&~@skipOnOcV11.0.0&&@api&&~@skip`
because there is no `-BEHAT_SUITE=${BEHAT_SUITE}` set in the `.drone.yml` in core the setting is not picked up in a manual run

sorry for the noice

phil-davis:
Today's "standard" CI-script changes have been made, like reference to ``tests/output/*.xml`` and content of ``.gitgnore``. The previously long list of jumbled commits has been squashed. The scripts here should match the "new current style standard"

@patrickjahns @DeepDiver1975 ready for review again.

phil-davis:
Rebased and resolved conflicts.
@patrickjahns @DeepDiver1975 please review. I like to be able to finish things.

PR #288, opened 2019-01-16
Do not scan files if etag hasn't changed

### LINKED ISSUE #218: Track file modifications in bg scanner
There is no need to rescan files if the content hasn't changed.
We'd better have an attribution to the file content using `eTag` for bg scanner ad trigger rescan if `eTag` of the avir differs from fiile `etag`.
In addition there should be a filter on a scantime in order not to stuck with the same set of frequently changed files.

VicDeo:
Closes #218 
- [x] filter files for scan by matching etags
- [x] sort files for scan by the previous scan time asc so frequently changed files will never be on the top of the list
- [x] upsert for successful background scan instead of legacy OC_DB delete+insert
- [x] use dependency injection where possible


lib/Cron/Task.php
```diff
@@ -32,12 +109,116 @@ public function __construct() {
...
+		if (!\OCP\App::isEnabled('files_antivirus')
+			|| $this->appConfig->getAvScanBackground() !== 'true'
+		) {
+			return;
+		}
+
+		// locate files that are not checked yet
+		try {
+			$result = $this->getFilesForScan();
+		} catch (\Exception $e) {
```
jvillafanez: I guess the exceptions here will be mostly from the DB layer... I'm not sure if this is one of those cases where catching every exception is fine

lib/Cron/Task.php
```diff
@@ -32,12 +109,116 @@ public function __construct() {
...
+				['app' => 'files_antivirus']
+			);
 			return;
 		}
 
-		$application = new Application();
-		$container = $application->getContainer();
-		$container->query('BackgroundScanner')->run();
+		$cnt = 0;
+		while (($row = $result->fetch()) && $cnt < self::BATCH_SIZE) {
```
jvillafanez: Since the usual way is `while (($row = $result->fetch()) !== false)`, I guess that in this case it isn't possible the fetch a value that could be misinterpreted a false.

lib/Cron/Task.php
```diff
@@ -32,12 +109,116 @@ public function __construct() {
...
+		}
+		return false;
+	}
+
+	/**
+	 * @param \OCP\IUser $user
+	 *
+	 * @return \OCP\Files\Folder
+	 */
+	protected function getUserFolder(IUser $user) {
```
jvillafanez: This function doesn't seems to be used

lib/Db/FileCollection.php
```diff
@@ -0,0 +1,102 @@
...
+
+	/**
+	 * @param int $fileSizeLimit
+	 *
+	 * @return \Doctrine\DBAL\Driver\Statement|int
+	 */
+	public function getCollection($fileSizeLimit) {
+		$dirMimeTypeId = $this->getDirectoryMimeTypeId();
+		$qb = $this->dbConnection->getQueryBuilder();
+		if ($this->dbConnection->getDatabasePlatform() instanceof MySqlPlatform) {
```
jvillafanez: any better alternative? I'd prefer not to deal with DB specific stuff in any app if possible.

lib/Cron/Task.php
```diff
@@ -32,12 +109,116 @@ public function __construct() {
...
-		if (!\OCP\App::isEnabled('files_antivirus')) {
+		if (!\OCP\App::isEnabled('files_antivirus')
+			|| $this->appConfig->getAvScanBackground() !== 'true'
+		) {
+			return;
+		}
+
+		// locate files that are not checked yet
+		try {
+			$result = $this->getFilesForScan();
```
jvillafanez: It might be worthy to return the results sorted by user in order to reduce the number of times we need to switch the FS. Something to be checked later.

lib/Db/FileCollection.php
```diff
@@ -0,0 +1,102 @@
...
+			->orderBy('fa.check_time', 'ASC');
+
+		return $qb->execute();
+	}
+
+	/**
+	 * @return int
+	 */
+	protected function getDirectoryMimeTypeId() {
+		return \OC::$server->getMimeTypeLoader()->getId(
```
jvillafanez: Inject the MimeTypeLoader?

lib/Item.php
```diff
@@ -190,23 +198,36 @@ public function processUnchecked(Status $status, $isBackground) {
...
-			$stmt = \OCP\DB::prepare('DELETE FROM `*PREFIX*files_antivirus` WHERE `fileid` = ?');
-			$result = $stmt->execute([$this->id]);
-			if (\OCP\DB::isError($result)) {
-				//TODO: Use logger
-				$this->logError(__METHOD__. ', DB error: ' . \OCP\DB::getErrorMessage());
-			}
-			$stmt = \OCP\DB::prepare(
-				'INSERT INTO `*PREFIX*files_antivirus` (`fileid`, `check_time`) VALUES (?, ?)'
+			$dbConnection = \OC::$server->getDatabaseConnection();
+			$delete = $dbConnection->getQueryBuilder();
```
jvillafanez: Why use delete + insert instead of just update the row?. If the expectation is that the file will likely exists (the cases where the file will be missing are quite a few), you might want to try an update and if nothing updates then insert. This should perform better since most of the time you'll hit the DB once instead of twice.

appinfo/Migrations/Version20180116141548.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+namespace OCA\Files_Antivirus\Migrations;
+use OCP\IDBConnection;
+use OCP\Migration\ISqlMigration;
+
+/**
+ * Cleans table before adding etag field
+ */
+class Version20180116141548 implements ISqlMigration {
```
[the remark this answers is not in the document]
VicDeo: > Unless this data are useless. 
Existing data is file id and scan time. Due to empty etags all existing entries are subject for rescan anyway. The next migration will be faster if we truncate the table.

lib/Db/FileCollection.php
VicDeo: not possible.

VicDeo:
@jvillafanez  Please note, that I didn't write the code in Cron/Task but just moved it from BackgroundScanner class since it is unneeded.

lib/Cron/Task.php
VicDeo: argh. it is not

lib/Db/FileCollection.php
VicDeo: no alternative until the core mounts are changed. I need join strings and it is 'CONCAT' for MySQL only. For all other DB engines it will be `||`

lib/Item.php
VicDeo: As for now I just rewrote the queries to use query builder. But if you don't like it I can revert this change to call `OC_DB` ;)

lib/Cron/Task.php
VicDeo: removed

lib/Db/FileCollection.php
VicDeo: done

appinfo/Migrations/Version20180116141548.php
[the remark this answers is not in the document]
jvillafanez: Ok, then I think it's better to merge both migrations into one and include a comment explaining why we're removing all the entries.

lib/Item.php
jvillafanez: I'm fine if this change is done later as long as there is a ticket for that.

appinfo/Migrations/Version20180116141548.php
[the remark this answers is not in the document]
VicDeo: We have ISchemaMigration and ISqlMigration interfaces. ISchemaMigration is not intended to truncate tables and ISqlMigration is not intended to change schema.

Currently ISqlMigration is executed before ISchemaMigration and it is possible to have a single migration implementing both interfaces but in general the order is undefined.

lib/Cron/Task.php
VicDeo: it's an 3rd party application cron job. Any exception inside it is probably related to the app and should be logged by the app itself.

appinfo/Migrations/Version20180116141548.php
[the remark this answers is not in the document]
jvillafanez: I misread the interfaces...
Any chance to check the performance loss of the migration if we don't clean the table? I guess we expect millions of rows in the table, not just a couple of hundreds.

lib/Item.php
VicDeo: np, we have upsert in 10.0.9. This part has been rewritten to use it.

lib/Cron/Task.php
VicDeo: Previous scan time should have precedence anyway as cron could stuck on the set of frequently changed files otherwise.

VicDeo:
@jvillafanez thanks. I just finished retesting recent changes. Everything works as expected.

davitol:
> @jvillafanez thanks. I just finished retesting recent changes. Everything works as expected.

@VicDeo Can you please provide the testing steps for this PR? THX

VicDeo:
### Update
updating the app removes all existing entries for oc_files_antivirus table

### Cron
Cron execution scans some files and put fileId, scan_time and etag into oc_files_antivirus table
it is done in batches, 10 files per one cron execution. 
 the order should be the following: 
  - not scanned files (that have no entries in oc_files_antivirus) have greater priority
 - already scanned files ordered be scan_time ascending **but** they shouldn't be scanned if etag is the same with oc_filecache

I compared entries and scan_time in oc_files_antivirus before and after cron execution to test this.

So the testing could include the following cases:
### Test Case 1 - Update
1. install and setup the prev version of the app
2. enable it
3. upload some files, **e.g. 15** and run cron.php
4. check that oc_files_antivirus has 10 records
5. Update to the app to the new version.
6. check that oc_files_antivirus has no entries anymore

### Case 2 - Cron execution
1. run cron.php on the updated app
2. check that oc_files_antivirus has got 10 records
3. manually update oc_files_antivirus to have different etags for e.g. 5 records
4. run cron.php
5. check  oc_files_antivirus table. 5 more records should appear (given that 15 files were uploaded in Case 1 p.3)
6. Check that entries updated manually in Case 2 p3 have got original etags and updated scan_time.
7 Check that any scan_time for 5 new records from Case2 p5 is less than scan_time for  the entries from Case 2 p3. (Which means that not scanned files always have greater priority)
8 Check that unmodified entries has the same scan_time as before the cron execution

VicDeo:
P.S. cron job is configured to be executed once per 15 minutes so you'll need to execute the following query to run it more frequently:
`UPDATE oc_jobs SET last_run=0,last_checked=0,reserved_at=0 where class='OCA\Files_Antivirus\Cron\Task';`

Issue #295, opened 2019-02-15
Acceptance test antivirusFileSize.feature:45 is failing


phil-davis:
See https://drone.owncloud.com/owncloud/files_antivirus/550/262
```
--- Failed scenarios:

    /var/www/owncloud/apps/files_antivirus/tests/acceptance/features/apiAntivirus/antivirusFileSize.feature:45

26 scenarios (25 passed, 1 failed)
255 steps (253 passed, 1 failed, 1 skipped)
6m25.51s (16.13Mb)
```
and
```
  Scenario: Files smaller than the upload threshold are checked for viruses when using chunking (use async move to upload)           # /var/www/owncloud/apps/files_antivirus/tests/acceptance/features/apiAntivirus/antivirusFileSize.feature:45
    Given parameter "av_max_file_size" of app "files_antivirus" has been set to "100"                                                # FeatureContext::serverParameterHasBeenSetTo()
    And using new DAV path                                                                                                           # FeatureContext::usingOldOrNewDavPath()
    And the administrator has enabled async operations                                                                               # FeatureContext::triggerAsyncUpload()
    When user "user0" uploads the following chunks asynchronously to "/myChunkedFile.txt" with new chunking and using the WebDAV API # FeatureContext::userUploadsTheFollowingChunksUsingNewChunking()
      | 1 | X5O!P%@AP[4\PZX54(P^)7C |
      | 2 | C)7}$EICAR-STANDARD-ANT |
      | 3 | IVIRUS-TEST-FILE!$H+H*  |
    Then the HTTP status code should be "202"                                                                                        # FeatureContext::theHTTPStatusCodeShouldBe()
    And the oc job status values of last request for user "user0" should match these regular expressions                             # FeatureContext::jobStatusValuesShouldMatchRegEx()
      | status | /^error$/ |
    And the last lines of the log file should contain log-entries containing these attributes:                                       # AntivirusContext::theLastLinesOfTheLogFileShouldContainEntriesWithTheseAttributes()
      | user  | app               | message               |
      | user0 | files_antivirus   | Infected file deleted |
      | user0 | no app in context | Exception             |
      log entry:
[...]

PR #296, opened 2019-02-18
Add extra expected log file exception message for antivirusFileSize.feature:45

phil-davis:
Issue #295 
When uploading with chunking, there is now the usual "Infected file deleted" message, but there are 2 exception messages logged. So we need to list both of those, so that the more interesting "Infected file deleted" message is found as the 3rd-last message in the log file.


phil-davis:
The core test step:
```
Then the last lines of the log file should contain log-entries containing these attributes:
```
is very fussy. The log file entries have to be in exactly the given order, and there have to be no other entries between or after. That makes tests quite fragile.

@individual-it do we try to make the core step more flexible for the future?

phil-davis:
drone failed in https://drone.owncloud.com/owncloud/files_antivirus/555/240 prepare-objectstore step.
All the tests related to this PR passed.
I will restart drone.

PR #315, opened 2019-07-05
Test app against latest release tarball

skshetry:
Add drone matrices to run test against latest stable release tarball.

Related Issue https://github.com/owncloud/QA/issues/611


skshetry:
Closing this as #316 solves the problem for now. We could revisit this later when we try to shift to drone 1.0.

PR #321, opened 2019-09-02
Initial drone starlark

phil-davis:
Special stuff:

- [x] clamav environment for "unit" tests - the test setup code has been modified so that it will use `AVIR_HOST` if that env var is defined.
- [x] activity app install bits
- [x] objectstore bits

See https://github.com/owncloud/activity/pull/782 to see the extra code added to support the various things that happen for tests here.

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -1,0 +1,1010 @@
+config = {
+	'app': 'files_antivirus',
+	'rocketchat': 'builds',
+
+	'branches': [
+		'master'
+	],
+
+	'appInstallCommand': 'make',
+
+	'codestyle': {
+		'ordinary' : {
+			'phpVersions': [
+				'7.0',
+				'7.1',
+				'7.2',
+				'7.3',
+			],
+		},
+	},
+
+	'phpunit': {
+		'allDatabases' : {
+			'phpVersions': [
+				'7.0',
+			]
+		},
+		'reducedDatabases' : {
+			'phpVersions': [
+				'7.1',
+				'7.2',
+				'7.3',
+			],
+			'databases': [
+				'sqlite',
+				'mariadb:10.2',
+			],
+			'coverage': False
+		},
+	},
+
+	'acceptance': {
+		'webUI': {
+			'suites': {
+				'webUIActivityList': 'webUIActList',
+			},
+			'browsers': [
+				'chrome',
+				'firefox'
+			],
+		},
+		'api': {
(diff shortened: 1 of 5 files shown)


.drone.starlark
```diff
@@ -0,0 +1,1073 @@
...
+				'mariadb:10.2',
+				'mysql:5.5',
+				'mysql:5.7',
+				'postgres:9.4',
+				'oracle'
+			],
+		},
+		'api-objectstore': {
+			'suites': {
+				'apiAntivirus': 'apiAntivirusObjectstore',
```
phil-davis: The name `apiAntivirusObjectstore` gets used as part of the matrix entry name in the drone.
So having a unique name here makes the drone matrix entry names unique.
Otherwise you end up with duplicate matrix entry names like `apiAntivirus-master-mariadb10.2-php7.0`

.drone.starlark
```diff
@@ -0,0 +1,1073 @@
...
+			},
+			'databases': [
+				'mariadb:10.2',
+			],
+			'extraApps': {
+				'files_primary_s3': 'composer install'
```
phil-davis: `files_primary_s3` needs at least `composer install` to get it ready to use.
(that is what the previous drone did to make it work)

.drone.starlark
```diff
@@ -0,0 +1,1073 @@
...
+			},
+			'extraSetup': {
+				'name': 'configure-app',
+				'image': 'owncloudci/php:7.0',
+				'pull': 'always',
+				'commands': [
```
phil-davis: For testing with objectstore we need to both:
1) setup the scality config for files_primary_s3, and;
2) setup the "usual" antivirus settings

.drone.starlark
```diff
@@ -0,0 +1,1073 @@
...
+					'image': 'owncloudci/php:7.0',
+					'pull': 'always',
+					'commands': [
+						# dummy clamav required for unit tests
+						'php tests/util/avirserver.php',
+					]
+				}
+			],
+			'extraEnvironment': {
+				'AVIR_HOST': 'dummy-clamav'
```
phil-davis: The "unit" tests do not mock everything. They depend on having an antivirus service.
The unit test code has been modified so, instead of being hard-coded to `0.0.0.0` or `127.0.0.1`, it will connect with `AVIR_HOST` if that environment variable is defined.

PR #324, opened 2019-09-26
Run API acceptance tests on both scality and ceph

phil-davis:
Some app repos test objectstore with scality and some with ceph.
Actually we can be more sure, and run the tests on both.

Also, `files_primary_s3` `master` has been bumped to require core min version `10.3`. So when testing with `files_primary_s3` we cannot use core `latest` = `10.2.1` any more. So  that has also been adjusted to use `10.3.0alpha2`.

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -67,10 +67,14 @@ config = {
 				'oracle'
 			],
 		},
-		'api-objectstore': {
+		'api-scality': {
 			'suites': {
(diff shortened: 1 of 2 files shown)


phil-davis:
`ceph` gets problems. I will look...
The `scality` against `10.3.0alpha2` stufff is more essential to do. That is now in PR #325

phil-davis:
See PR #326

PR #330, opened 2019-10-08
Update drone starlark with latest code 20191011

phil-davis:
Update to "latest" drone starlark app code:
1) support running all suites split into a number of parts (some apps run core tests like this)
2) put knowledge about the `ceph` and `scality` S3 backends into the code, to remove repetition of this knowledge in multiple app configs
3) added a `build()` pipeline (that was used by the testing app, and might be useful in future for other apps?)

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -51,9 +51,9 @@ config = {
 				'chrome',
 				'firefox'
 			],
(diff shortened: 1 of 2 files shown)


individual-it:
CI failed because of drone timeouts, restarted

PR #337, opened 2019-11-28
Drop PHP 7.0 from drone CI

phil-davis:
Because core has dropped PHP 7.0 support https://github.com/owncloud/core/pull/36290


phil-davis:
I will have to override the codecov later - need to go now

PR #346, opened 2019-12-23
[DO NOT MERGE] Demonstration that #347 fixes the problem

skshetry:


tests/acceptance/wait-for-stats-clamav.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
```
skshetry: I don't think we need this script, as all builds are returing successfully in 0seconds. Done in 4bee2b030fd9960885ed9, if needed in future.

Issue #340, opened 2019-12-12
test antivirusFileSize.feature:13 fails


phil-davis:
Last night: https://drone.owncloud.com/owncloud/files_antivirus/1038/24/12
and https://drone.owncloud.com/owncloud/files_antivirus/1038/19/12
```
--- Failed scenarios:

    /var/www/owncloud/testrunner/apps/files_antivirus/tests/acceptance/features/apiAntivirus/antivirusFileSize.feature:13

29 scenarios (28 passed, 1 failed)
295 steps (292 passed, 1 failed, 2 skipped)
```

```
  Scenario: Files smaller than the upload threshold are checked for viruses                                                 # /var/www/owncloud/testrunner/apps/files_antivirus/tests/acceptance/features/apiAntivirus/antivirusFileSize.feature:13
    Given parameter "av_max_file_size" of app "files_antivirus" has been set to "100"                                       # FeatureContext::serverParameterHasBeenSetTo()
    When user "user0" uploads file "eicar.com" from the antivirus test data folder to "/virusfile.txt" using the WebDAV API # AntivirusContext::userUploadsFileFromAntivirusDataFolderTo()
    Then the HTTP status code should be "403"                                                                               # FeatureContext::theHTTPStatusCodeShouldBe()
      HTTP status code is not the expected value
      Failed asserting that 201 matches expected '403'.
    And the last lines of the log file should contain log-entries containing these attributes:                              # AntivirusContext::theLastLinesOfTheLogFileShouldContainEntriesWithTheseAttributes()
      | user  | app             | method | message               |
      | user0 | files_antivirus | PUT    | Infected file deleted |
    And as "user0" file "/virusfile.txt" should not exist                                                                   # FeatureContext::asFileOrFolderShouldNotExist()
```

phil-davis:
@HariBhandari07 this started failing last night. Please assign someone to investigate.

HariBhandari07:
@phil-davis I have run the failing scenario multiple times in this PR https://github.com/owncloud/files_antivirus/pull/341 and it passes. Should we close this issue and just declare it flakey.

phil-davis:
Maybe this is a timing issue and in drone we need to wait a bit until the AV service is up-and-running.
(something similar to what we did for the elastic search service)
We could put in a `wait-for-it` anyway - just need to work out what exactly to wait for ;)

phil-davis:
[...]

PR #347, opened 2019-12-24
Change docker image which waits for virus db update, increase wait

skshetry:
I have changed the docker image to be `skshetry/clamavd`, which is available in https://github.com/JankariTech/docker-clamavd/pull/1, and increased the wait time to 5 minutes (doesn't really hurt).

~I also noticed, that there is a certain time for clamavd to get ready. So, I added a script, so, as to wait for   `clamavd` to respond for `nSTATS` command~.

~This should make it more reliable~.

Related issue #340

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -84,7 +84,7 @@ config = {
 					'pull': 'always',
 					'commands': [
 						'cd /var/www/owncloud/server',
-						'wait-for-it -t 60 clamav:3310',
+						'wait-for-it -t 300 clamav:3310',
 						'php occ config:app:set --value "clamav" files_antivirus av_host',
 						'php occ config:app:set --value "daemon" files_antivirus av_mode',
 						'php occ config:app:set --value "3310" files_antivirus av_port',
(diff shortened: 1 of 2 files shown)


phil-davis:
`owncloudci/php` images does not have `nc` installed. https://drone.owncloud.com/owncloud/files_antivirus/1077/28/13

skshetry:
The script can be found in the 4bee2b0, if needed in future. I don't think we need this script at the moment, as all builds are returning successfully in 0seconds. But, if it fails for some reason, it might be the timing issue with `clamav` in that it hasn't really started accepting streams. Hopefully, the script can help in such case.

skshetry:
@phil-davis, the tests have passed here. I also made a demo PR #346 so to run the tests multiple times and it has passed as well(https://drone.owncloud.com/owncloud/files_antivirus/1085/50/1). Can you please review and merge?

HariBhandari07:
LGTM

skshetry:
@phil-davis, I have used our latest organizational docker image. Please review again.

PR #351, opened 2019-12-31
Ci debug

dpakach:
just a trigger to check failing tests.


dpakach:
Found Problem and fixed by https://github.com/owncloud/core/pull/36668

PR #354, opened 2020-02-08
Update starlark code 20200208

phil-davis:
Issue https://github.com/owncloud/QA/issues/640

Also fixes issue #356 (I had to cherry-pick the fix for that from PR #357 - everything needs to be together to get CI passing)


phil-davis:
Getting an unexpected difference in the virus detection messages.
Issue #356

Issue #356, opened 2020-02-09
Virus infection message is different to expected


phil-davis:
https://drone.owncloud.com/owncloud/files_antivirus/1152/14/14
```
--- Failed scenarios:

    /var/www/owncloud/testrunner/apps/files_antivirus/tests/acceptance/features/webUIActivityList/activityList.feature:23
    /var/www/owncloud/testrunner/apps/files_antivirus/tests/acceptance/features/webUIActivityList/activityList.feature:24
    /var/www/owncloud/testrunner/apps/files_antivirus/tests/acceptance/features/webUIActivityList/activityList.feature:39
    /var/www/owncloud/testrunner/apps/files_antivirus/tests/acceptance/features/webUIActivityList/activityList.feature:40

5 scenarios (1 passed, 4 failed)
57 steps (35 passed, 4 failed, 18 skipped)
```

```
  Scenario Outline: replacing a file by a virus file should be listed in the general activity list                # /var/www/owncloud/testrunner/apps/files_antivirus/tests/acceptance/features/webUIActivityList/activityList.feature:26
    Given using <dav-path-version> DAV path                                                                       # FeatureContext::usingOldOrNewDavPath()
    And user "user1" has uploaded file "eicar.com" from the antivirus test data folder to "lorem.txt"             # AntivirusContext::userUploadsFileFromAntivirusDataFolderTo()
    When the user browses to the activity page                                                                    # WebUIActivityContext::userHasBrowsedToTheActivityPage()
    Then the activity number 1 should contain message "File files/lorem.txt" in the activity page                 # WebUIActivityContext::theActivityNumberShouldContainMessageInTheActivityPage()
    And the activity number 1 should contain message "is infected with Eicar-Test-Signature" in the activity page # WebUIActivityContext::theActivityNumberShouldContainMessageInTheActivityPage()
    And the activity should not have any message with keyword "change"                                            # WebUIActivityContext::theActivityShouldNotHaveAnyMessageWithKeyword()
    When the user filters activity list by "Antivirus"                                                            # WebUIActivityContext::theUserFiltersActivityListBy()
    Then the activity number 1 should contain message "File files/lorem.txt" in the activity page                 # WebUIActivityContext::theActivityNumberShouldContainMessageInTheActivityPage()
    And the activity number 1 should contain message "is infected with Eicar-Test-Signature" in the activity page # WebUIActivityContext::theActivityNumberShouldContainMessageInTheActivityPage()
[...]

PR #357, opened 2020-02-10
Adjust expected activity message for infected file

phil-davis:
Fixes issue #356 

CI started reporting `is infected with Clamav.Test.File-7` instead of `is infected with Eicar-Test-Signature`. Actually the clamav file `Clamav.Test.File-7` is the Eicar test file. In CI the Clamav database is automagically update before the CI run, so we get an up-to-date database each CI  run. That means that Clamav could change the detail of what they report as the "virus" that has been detected.

For these tests, we only care that some virus has been detected, not the exact name of the "virus".

Make the test expectation not depend on the exact name of the virus.


phil-davis:
This has the Oracle timeout problem - needs #354 first. But #354 needs this first - catch-22.
I have cherry-picked this into #354

PR #360, opened 2020-02-12
[Tests-only] Run core tests with files_antivirus enabled

### LINKED ISSUE #359: Run Core acceptance tests with files_antivirus
Part of owncloud/QA#639

Run all or some of the core acceptance tests with files_antivirus.
These tests can be run only on the nightly runs.

dpakach:
Fixes #359 

Run core acceptance tests with files_antivirus

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -133,6 +133,63 @@ config = {
 				},
 			]
 		},
+		'core-api-acceptance': {
+			'suites': [
+				'apiAll',
+			],
+			'databases': [
+				'mysql:5.7',
+			],
+			'servers': [
+				'daily-master-qa',
+			],
+			'phpVersions': [
+				'7.1',
+			],
+			'runCoreTests': True,
+			'federatedServerNeeded': True,
+			# 'cron': 'nightly',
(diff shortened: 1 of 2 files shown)


dpakach:
5d0d4a103df51e6356941d22bebadcb3c6e1118e Shows all core tests passing in CI.

.drone.star
```diff
@@ -133,6 +133,63 @@ config = {
 				},
 			]
 		},
+		'core-api-acceptance': {
+			'suites': [
+				'apiAll',
```
kiranparajuli589: where does this come form?
How it works? (JFI)

dpakach: Seems like a magic word that just brings in all api tests from Core. Maybe @phil-davis  knows more about it.

phil-davis: when you specify `'runCoreTests': True` then the starlark creates `numberOfParts` pipelines and names each pipeline like `apiAll-35-1-master-7.2`

`run.sh` in core knows how to select the suites from all the `api*` suites in core. If needed it will run multiple suites in a single pipeline...

PR #361, opened 2020-02-24
[Tests-only] guzzle version updated

HariBhandari07:
### Description
This PR updates guzzle version to 6.5

### Related Issue
https://github.com/owncloud/core/pull/36999


individual-it:
`        Fatal error: Class 'GuzzleHttp\Stream\Stream' not found (Behat\Testwork\Call\Exception\FatalThrowableError)`

PR #373, opened 2020-07-27
Objectstore implementations do no use part files - as a result Permis…

DeepDiver1975:
…sionsMask wrapper cannot delete the infected file. To bypass this antivirus will ignore all wrappers and delete directly on the physical storage.

## To do
- [ ] manual test in the bayernbox container @enbrnz 
- [ ] needs release ticket @HanaGemela


micbar:
@enbrnz Could you do the manual test?

enbrnz:
I was able to confirm that the patch works:
1. Spun up an environment without patch 
2. Uploaded test virus through public link
3. Installed patch
4. No longer able to upload test virus through public link
5. Background job deleted test virus, that was uploaded in step 2.

PR #379, opened 2020-11-02
Allow to use a scanner class living outside the files_antivirus app

DeepDiver1975:


phil-davis:
Note: issue https://github.com/owncloud/core/issues/38067
https://drone.owncloud.com/owncloud/files_antivirus/1506/14/14

dependencies are broken due to a composer 2.0 problem/feature. I will make a PR now to this repo...

phil-davis:
#380 has been merged. @DeepDiver1975 Please rebase.

lib/ScannerFactory.php
```diff
@@ -54,6 +54,11 @@ public function __construct(AppConfig $appConfig, ILogger $logger) {
 	 * @throws InitException
 	 */
 	protected function getScannerClass() {
+		$class = $this->appConfig->getExternalScannerClass();
```
jvillafanez, line 57: We might want to include a code comment to ensure we want to prioritize the external scanner over whatever is configured inside the app. We could use another option -> daemon, socket, executable, configured_in_system_config, for example.

DeepDiver1975: Config section has been completly disabled

Screenshot from 2020-11-09 14-42-10

lib/Status.php
```diff
@@ -29,24 +29,31 @@ class Status {
...
 	protected $ruleMapper;
-	
+
 	public function __construct() {
 		$this->numericStatus = self::SCANRESULT_UNCHECKED;
 		$this->ruleMapper = new Db\RuleMapper(\OC::$server->getDb());
 	}
-	
+
+	public static function create(int $status, string $details = '') {
```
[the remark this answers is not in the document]
jvillafanez: I guess the expected usage for the class is to create the `Status` object, and then run `$status->parseResponse(...)`. This is intended to set the proper status and details based on the output response of the virus scanner. In this regard, this new `create` method is risky because you could bypass those checks.
Note that the method is currently unused within this repo.

My main worry is that, after half a year we think: 1) the method is risky because it can be misused, and 2) the method isn't used. As a consequence, we remove the method, and then we break an app that we hadn't considered.
If the method is required for a 3rd-party app, we need to make sure that dependency is visible, specially when the files_antivirus doesn't have an API for this matter.

jvillafanez (approved):

PR #380, opened 2020-11-02
Adjust PHP dependencies for composer 2.0

phil-davis:
Issue https://github.com/owncloud/core/issues/38067

See issue for details - `behat` acceptance tests need to use PHP  7.4 for the dependencies to sort themselves out with composer 2.0

And adjusting owncloud min-version="10.3" because we want to ensure PHP 7.1 and... - it would require a bunch of extra testing to ensure that we could run the latest encryption app code with ownCloud 10.2 and PHP 7.0

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1052,7 +1052,7 @@ def acceptance():
 									},
 									'steps':
 										installCore(server, db, params['useBundledApp']) +
-										installTestrunner(phpVersion, params['useBundledApp']) +
+										installTestrunner('7.4', params['useBundledApp']) +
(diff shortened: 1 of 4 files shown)


dpakach (approved):

phil-davis:
This uses files_primary_s3 to do some acceptance test pipelines with S3 storage.
Needs https://github.com/owncloud/files_primary_s3/pull/395 merged first, then rerun CI.

phil-davis:
Restarted drone CI - let's see the result...

Issue #394, opened 2021-02-05
[QA] occ app:check fails


jnweiger:
Seen with files_antivirus-0.16.0RC1  installed in oc-10.6.0:

`occ app:check files_antivirus`
```
Analysing /home/var.www/owncloud/apps-external/files_antivirus/lib/Item.php
 1 errors
    line  178: OCP\JSON - Static method of deprecated class must not be called
Analysing /home/var.www/owncloud/apps-external/files_antivirus/lib/Cron/Task.php
 1 errors
    line  212: OC_Util - Static method of private class must not be called
Analysing /home/var.www/owncloud/apps-external/files_antivirus/appinfo/preupdate.php
 3 errors
    line    8: OCP\DB - Static method of deprecated class must not be called
    line   13: OCP\DB - Static method of deprecated class must not be called
    line   23: OCP\DB - Static method of deprecated class must not be called
Deprecated file found: /home/var.www/owncloud/apps-external/files_antivirus/appinfo/update.php - please use repair steps
App is not compliant
```

VicDeo:
The first  needs quite lot of efforts to eliminate it
The second is not fixable at all with current core -  there is no method to teardown an user filesystem within a public namespace.

All the rest are addressed in https://github.com/owncloud/files_antivirus/pull/398

PR #398, opened 2021-02-12
Get rid of  update.php and preupdate.php

VicDeo:
### some notes:
 - Rules are already populated in https://github.com/owncloud/files_antivirus/blob/master/appinfo/Migrations/Version20170808222334.php 
so we can ignore the population part of `update.php`
- files_antivirus table is truncated in https://github.com/owncloud/files_antivirus/blob/master/appinfo/Migrations/Version20180116141548.php so we can ignore the deduplication part of `preupdate.php`

Partially addresses https://github.com/owncloud/files_antivirus/issues/394


jvillafanez (approved):
Assuming it works without problems

Issue #396, opened 2021-02-05
[QA] default mode 'executable' is slowest


jnweiger:
Seen with files_antivirus 0.16.0RC1 in oc-10.6.0

Following the setup instructions in https://doc.owncloud.com/server/10.6/admin_manual/configuration/server/virus-scanner-support.html
The documentation discusses mode configuration.
For mode 'Executable' it warns:
"TIP
clamscan is slow and not always reliable for on-demand usage; it is better to use one of the daemon modes."

Nevertheless, the installation defaults to mode Executable.
The clamscan binary indeed takes ca 20 seconds on a small file and uses ca 1GB of RAM.
In daemon mode a small file is processed instantaneously.

Expected behaviour: Use mode "Daemon (socket)" mode per default.

VicDeo:
@jnweiger the main idea behind this behavior is that `clamscan` is available in 99% of set ups while daemon/socket mode could be unavailable. So enabling the app makes the protection working right away.

There is a room for improvement definitely...
E.g. we can probe the socket/daemon defaults  () after the app is enabled  and use executable mode if the probe was unsuccessful

VicDeo:
https://github.com/owncloud/files_antivirus/pull/399 introduces autoprobe when the app is enabled for the first time.
Now `executable`  is used by default only after both  daemon and socket mode are proved not to respond

jnweiger:
Cool. That is good news.
We can accept a slow default for a while. Robustness is more important.

jnweiger:
Confirmed fixed in 1.0.0-rc1

PR #399, opened 2021-02-15
Prefer daemon or socket to executable mode to if any of those is available

VicDeo:
After the app is enabled **for the first time** we check the default port and default socket. If clamav detects eicar using any of those we use this mode as a default. Otherwise  the executable mode is used as before.

Note: This is done **only once** - after the app is enabled **for the first time.**

Closes https://github.com/owncloud/files_antivirus/issues/396

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -7,7 +7,7 @@
  *
  * @author Viktar Dubiniuk <dubiniuk@owncloud.com>
  *
- * @copyright Viktar Dubiniuk 2014-2018
+ * @copyright Viktar Dubiniuk 2021
  * @license AGPL-3.0
  */
 
@@ -28,7 +28,7 @@ use OCP\IL10N;
 class Application extends App {
 	public function __construct(array $urlParams = []) {
 		parent::__construct('files_antivirus', $urlParams);
-		
+
 		$container = $this->getContainer();
 		$container->registerService(
 			'RuleController',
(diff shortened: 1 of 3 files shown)


lib/AppInfo/Application.php
```diff
@@ -172,4 +172,25 @@ function ($mountPoint, $storage) {
...
+			if ($scannerFactory->testConnection($appConfig) === true) {
+				$set = true;
+				break;
+			}
+		}
+
+		if ($set === false) {
+			$appConfig->setAvMode('executable');
+		}
+		\OC::$server->getConfig()->setAppValue('files_antivirus', 'autoprobe', true);
```
jvillafanez, line 194: I think we can get it from the container.

jvillafanez (approved):
Just a tiny detail.

jvillafanez:
I'd prefer to have a "test configuration" button on the web UI so the admin can test. The problem with the autoprobe is that you have to guess where everything is, but if the guess is wrong it's quite useless.
In addition, I'm not sure if it's really recommended to have the antivirus in the same machine. From a security point of view, I guess so since you can have the clamd daemon on a unix socket, but it might increase the load on the server machine, which might not be desired.

lib/AppInfo/Application.php
VicDeo: updated

VicDeo:
> I'd prefer to have a "test configuration" button on the web UI so the admin can test.

We already have configuration test and it is done on save.

The issue is quite different: executable mode is used as a default because it has more chances to work out-of-a-box. This PR is created to use better (faster) options in case they are available.

PR #404, opened 2021-03-10
Add docstring to the unlink method in the AvirWrapper

JammingBen:
...because the reasoning behind it is not clear.

Related to https://github.com/owncloud/enterprise/issues/4427

### DIFF
--- a/lib/AvirWrapper.php
+++ b/lib/AvirWrapper.php
@@ -253,7 +253,12 @@ class AvirWrapper extends Wrapper {
 	}
 
 	/**
-	 * @inheritDoc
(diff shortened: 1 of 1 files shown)


jvillafanez (approved):

lib/AvirWrapper.php
```diff
@@ -253,7 +253,12 @@ private function isScannableSize($path) {
 	}
 
 	/**
-	 * @inheritDoc
+	 * Objectstore implementations do no use part files, so PermissionsMask wrapper
```
VicDeo, line 256: "do no use part files" -> do not use part files

PR #426, opened 2021-04-22
Prevent from crashing on missing or expired license

VicDeo:
0. Prerequisite: have **NO** valid license
1. Set mode to `ICAP`

### Expected for Web UI
UI Feedback that it's not possible to set `ICAP` as a mode due to an invalid license

### Expected for CLI
Value is saved (there is no way to control that) but is not used

### Actual behavior (both CLI/Web UI)
OC is constantly crashing until a valid license is provided, the mode is changed to anything else but `ICAP`  or files_antivirus is disabled. All these actions are possible via CLI only as Web UI is not working any more.

## Changed in this PR
- No hard crash any more
- switching mode to `icap` via Web UI is prevented if the license is invalid
- in case the license is expired or mode was set to `icap` via CLI the mode value is overloaded with `executable` at the runtime.
The message 'No valid license found for icap scanner, resetting mode to executable' is written to log file accordingly.

### DIFF
--- a/lib/AppConfig.php
+++ b/lib/AppConfig.php
@@ -164,6 +164,8 @@ class AppConfig {
 	 * @param string $value
 	 *
 	 * @return void
+	 *
+	 * @throws \UnexpectedValueException
 	 */
 	public function setAppValue($key, $value) {
 		$this->validateValue($key, $value);
@@ -175,9 +177,15 @@ class AppConfig {
 	 * @param string $value
 	 *
 	 * @return void
+	 *
+	 * @throws \UnexpectedValueException
 	 */
 	public function validateValue($key, $value) {
-		if ($key === 'av_mode' && $value === 'icap'&& !$this->licenseManager->checkLicenseFor('icap')) {
+		if (
+			$key === 'av_mode'
+			&& $value === 'icap'
+			&& !$this->licenseManager->checkLicenseFor($this->appName, ["disableApp" => false])
+		) {
 			$this->logger->error('No valid license found for icap scanner');
 			throw new \UnexpectedValueException("No valid license found for icap scanner mode");
 		}
--- a/lib/AppConfig.php
+++ b/lib/AppConfig.php
@@ -14,6 +14,8 @@
 namespace OCA\Files_Antivirus;
 
 use \OCP\IConfig;
+use OCP\ILogger;
+use OCP\License\ILicenseManager;
 
 /**
  * @method string getAvMode()
@@ -51,6 +53,16 @@ class AppConfig {
 	 */
 	private $config;
 
+	/**
+	 * @var ILicenseManager
+	 */
+	private $licenseManager;
+
+	/**
+	 * @var ILogger
+	 */
+	private $logger;
+
 	private $defaults = [
 		'av_mode' => 'executable',
 		'av_socket' => '/var/run/clamav/clamd.ctl',
@@ -70,9 +82,13 @@ class AppConfig {
 	 * AppConfig constructor.
 	 *
 	 * @param IConfig $config
+	 * @param ILicenseManager $licenseManager
+	 * @param ILogger $logger
 	 */
-	public function __construct(IConfig $config) {
+	public function __construct(IConfig $config, ILicenseManager $licenseManager, ILogger $logger) {
 		$this->config = $config;
+		$this->licenseManager = $licenseManager;
+		$this->logger = $logger;
 	}
 
 	public function getAvChunkSize() {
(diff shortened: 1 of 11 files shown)


VicDeo:
@jnweiger  @pmaier1

pmaier1:
Nice, thanks :+1:

DeepDiver1975:
> * Prerequisite: have **NO** valid license
> 
> * Set mode to `ICAP`

if there is no license setup a trial license will be created by core.

I assume you mean that the give license did expire - right?

VicDeo:
>if there is no license setup a trial license will be created by core.

@DeepDiver1975 yep. The crash happens after the grace period is expired - 24 hours for 10.7.0

jvillafanez:
I didn't know the files_antivirus was checking already the license for the icap app.... This is a no-go for me: apps should care only for their own licenses, if any. Since the files_antivirus app is public, it shouldn't have to have a dependency with the licenseManager.

https://github.com/owncloud/core/blob/eebd754771a99bc44820eb67a6d57cea238e6b86/lib/public/License/ILicenseManager.php#L115-L118

A better option could be for the files_antivirus app to trigger an event to register scanners for this particular option, and then check if the icap app is registered or not. To the files_antivirus app, the reason why the icap app isn't registered doesn't matter, that's something for the icap app to handle.
The license check would be performed by the icap app itself as initially intended.

For the short term, you can use `getLicenseStateFor('icap')` so you don't trigger an unwanted grace period or disable the icap app unintentionally, although I still don't like the idea too much.

To clarify a bit, I'm trying to be strict due to "partially public" apps such as richdocuments. The same scenario for richdocuments would cause the richdocuments app to get disabled even though it could work partially. My concern is that the piece of code in this PR would be taken as the way of doing things and would cause problems in the long run.

VicDeo:
@jvillafanez there is no icap any more. Currently 'icap' is an enterprise feature of the files_antivirus app. 
And TBH this is irrelevant to the topic of this PR that just prevents the system from hard crash for all users once the license is expired :)

jvillafanez:
Ok, if icap is an enterprise feature of the app that's fine, but then we should check for `checkLicenseFor('files_antivirus')` as originally intended. There are options documented to prevent core from disabling the app if the license expires, so you could still use the app although some features could be locked.

The `checkLicenseFor('icap')` will cause core to disable the icap app, which could cause problems because the app doesn't exists, or even it could be a valid app.

VicDeo:
> The checkLicenseFor('icap') will cause core to disable the icap app, which could cause problems because the app doesn't exists, or even it could be a valid app.

and do we  want to disable the  whole app due to enterprise config option involved (e.g. by mistake)?
please note that we are still discussing something that is not related to hard app fail which is fixed by this PR.

jvillafanez:
> and do we want to disable the whole app due to enterprise config option involved (e.g. by mistake)?

That's why we introduced the additional options in the `checkLicenseFor` method, so you can tell core to check the license but without disabling the app.
For this case, you can tell core to start a grace period if possible (in case there is no valid license) but don't disable the app. The method would return true while the grace period is active (enabling the icap feature), but it will return false **without disabling the app** once it's over (assuming there is no valid license)

VicDeo:
@jvillafanez thanks for clarification. Please rereview. I also bumped min OC version as `checkLicenseFor` options were introduced in 10.6.0

lib/AppConfig.php
```diff
@@ -140,10 +163,28 @@ public function getAppValue($key) {
 	 * @param string $key
 	 * @param string $value
 	 *
-	 * @return string
+	 * @return void
 	 */
 	public function setAppValue($key, $value) {
-		return $this->config->setAppValue($this->appName, $key, $value);
+		$this->validateValue($key, $value);
```
jvillafanez, line 169: This call throws an "UnexpectedValueException", so the method should either document it, or handle that particular exception

lib/AppConfig.php
```diff
@@ -140,10 +163,28 @@ public function getAppValue($key) {
...
+		$this->config->setAppValue($this->appName, $key, $value);
+	}
+
+	/**
+	 * @param string $key
+	 * @param string $value
+	 *
+	 * @return void
+	 */
+	public function validateValue($key, $value) {
```
jvillafanez, line 179: Any reason for this method to be public? If so, it should be documented that the method can throw an "UnexpectedValueException" under some conditions

lib/AppConfig.php
```diff
@@ -140,10 +163,28 @@ public function getAppValue($key) {
...
+		if (
+			$key === 'av_mode'
+			&& $value === 'icap'
```
jvillafanez, line 182: Quick style check: I'm more used to see the `&&` at the end of the condition. You can keep this if there are other places with the same style, so up to you.

lib/Controller/SettingsController.php
```diff
@@ -87,37 +87,47 @@ public function save($avMode, $avSocket, $avHost, $avPort,
...
 				'settings' => $this->settings->getAllValues()
-			]
-		);
+			];
+		} catch (\UnexpectedValueException $e) {
+			$response = [
+				'data' => [
+					'message' => $e->getMessage()
+				],
+				'connection' => 1,
```
jvillafanez, line 125: I'm worried this "1" value isn't documented anywhere...
Could I change it to "-1" or "999"? Does it have any meaning (such as 1 = cannot connect, 2 = test file failed to be detected, etc)

VicDeo: unrelated to the the issue

lib/AppConfig.php
VicDeo: `throws` has been added

lib/AppConfig.php
VicDeo: `throws` has been added. As it was called Initially  from a different class I decided do not modify the visibility  when I moved the usage into the same class.

jvillafanez (approved):

jnweiger:
Confirmed fixed in 1.0.0-rc2: The mode falls back to (really slow) executable. and cannot be switched to icap. Red error appears.
image

Issue #420, opened 2021-04-18
Icap unit tests fail sometimes


phil-davis:
https://drone.owncloud.com/owncloud/files_antivirus/1811/11/8
```
PHPUnit 8.5.15 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.4.15
Configuration: /var/www/owncloud/server/apps/files_antivirus/phpunit.xml
Error:         No code coverage driver is available

......................E.........                                  32 / 32 (100%)

Time: 6.64 seconds, Memory: 32.00 MB

There was 1 error:

1) OCA\Files_Antivirus\Tests\unit\Scanner\IcapScannerTest::testScannerAsync with data set "eicar" (1, 'X5O!P%@AP[4\PZX54(P^)7CC)7}$E...!$H+H*')
RuntimeException: Unknown ICAP response

/var/www/owncloud/server/apps/files_antivirus/lib/Scanner/ICAPClient.php:150
/var/www/owncloud/server/apps/files_antivirus/lib/Scanner/ICAPClient.php:120
/var/www/owncloud/server/apps/files_antivirus/lib/Scanner/ICAPScanner.php:45
/var/www/owncloud/server/apps/files_antivirus/tests/unit/Scanner/IcapScannerTest.php:53

ERRORS!
Tests: 32, Assertions: 52, Errors: 1.
```

https://drone.owncloud.com/owncloud/files_antivirus/1811/7/9
```
PHPUnit 8.5.15 by Sebastian Bergmann and contributors.

Runtime:       PHPDBG 7.2.34-16+ubuntu18.04.1+deb.sury.org+1
Configuration: /var/www/owncloud/server/apps/files_antivirus/phpunit.xml

......................EE........                                  32 / 32 (100%)

Time: 13.87 seconds, Memory: 54.52 MB

There were 2 errors:

1) OCA\Files_Antivirus\Tests\unit\Scanner\IcapScannerTest::testScannerAsync with data set "eicar" (1, 'X5O!P%@AP[4\PZX54(P^)7CC)7}$E...!$H+H*')
RuntimeException: Unknown ICAP response

/var/www/owncloud/server/apps/files_antivirus/lib/Scanner/ICAPClient.php:150
/var/www/owncloud/server/apps/files_antivirus/lib/Scanner/ICAPClient.php:120
/var/www/owncloud/server/apps/files_antivirus/lib/Scanner/ICAPScanner.php:45
[...]

PR #428, opened 2021-04-27
Do not depend on the sockets PHP extension

### LINKED ISSUE #427: ICAP implementation depends on the 'sockets' PHP  extension
I'd suggest to stay away from `socket_create` & co and migrate to `stream_socket_client` provided by PHP core package as it creates an additional dependency.

Otherwise we need to check whether `sockets` extension is loaded

VicDeo:
This PR makes no need to have sockets PHP extension installed and enabled. 
In most systems this extension is not  enabled out of a box. In this case files_antivirus is silently crashing after  'icap' mode is chosen.
Closes #427

### DIFF
--- a/lib/Scanner/ICAPClient.php
+++ b/lib/Scanner/ICAPClient.php
@@ -6,7 +6,7 @@ use RuntimeException;
 class ICAPClient {
 	private $host;
 	private $port;
-	private $socket;
+	private $writeHandle;
 
 	public $userAgent = 'ownCloud-icap-client/0.1.0';
 
@@ -15,28 +15,29 @@ class ICAPClient {
 		$this->port = $port;
 	}
 
-	private function connect(): bool {
-		$this->socket = \socket_create(AF_INET, SOCK_STREAM, SOL_TCP);
+	private function connect(): void {
+		$this->writeHandle = @\stream_socket_client(
+			"tcp://{$this->host}:{$this->port}",
+			$errorCode,
+			$errorMessage,
+			5
+		);
 
-		if (!\socket_connect($this->socket, $this->host, $this->port)) {
-			return false;
+		if (!$this->writeHandle) {
+			throw new InitException(
+				\sprintf(
+					'Cannot connect to "tcp://%s:%s": %s (code %d)',
+					$this->host,
+					$this->port,
+					$errorMessage,
+					$errorCode
+				)
+			);
 		}
-
-		# McAfee seems to not properly close the socket once all response bytes are sent to the client
-		# we use a 5 sec time out on receiving data
-		if (!\socket_set_option($this->socket, SOL_SOCKET, SO_RCVTIMEO, ["sec" => 5, "usec" => 0])) {
-			return false;
-		}
-		return true;
 	}
 
 	private function disconnect(): void {
-		@\socket_shutdown($this->socket);
-		@\socket_close($this->socket);
-	}
-
-	public function getLastSocketError(): int {
-		return \socket_last_error($this->socket);
+		@\fclose($this->writeHandle);
 	}
 
 	public function getRequest(string $method, string $service, array $body = [], array $headers = []): string {
@@ -121,14 +122,18 @@ class ICAPClient {
 	}
 
 	private function send(string $request): string {
-		if (!$this->connect()) {
-			throw new RuntimeException("Cannot connect to icap://{$this->host}:{$this->port} (Socket error: " . $this->getLastSocketError() . ")");
+		$this->connect();
+		if (@\fwrite($this->writeHandle, $request) === false) {
+			throw new InitException(
+				\sprintf('Writing to "%s:%s" failed', $this->host, $this->port)
+			);
 		}
 
-		\socket_write($this->socket, $request);
-
+		# McAfee seems to not properly close the socket once all response bytes are sent to the client
+		# we use a 5 sec time out on receiving data
+		\stream_set_timeout($this->writeHandle, 5, 0);
 		$response = '';
-		while ($buffer = \socket_read($this->socket, 2048)) {
+		while ($buffer = \fread($this->writeHandle, 2048)) {
 			$response .= $buffer;
 		}
 
--- a/lib/Scanner/ICAPClient.php
+++ b/lib/Scanner/ICAPClient.php
@@ -16,13 +16,13 @@ class ICAPClient {
 	}
 
 	private function connect(): void {
+		// Shut stupid uncontrolled messaging up - we handle errors on our own
 		$this->writeHandle = @\stream_socket_client(
 			"tcp://{$this->host}:{$this->port}",
 			$errorCode,
 			$errorMessage,
 			5
 		);
-
 		if (!$this->writeHandle) {
 			throw new InitException(
 				"Cannot connect to \"tcp://{$this->host}:{$this->port}\": $errorMessage (code $errorCode)"
@@ -31,6 +31,7 @@ class ICAPClient {
 	}
 
 	private function disconnect(): void {
+		// Due to suppressed output it could be a point of interest for debugging. Someday. Maybe.
 		@\fclose($this->writeHandle);
 	}
 
@@ -117,6 +118,7 @@ class ICAPClient {
 
 	private function send(string $request): string {
 		$this->connect();
+		// Shut stupid uncontrolled messaging up - we handle errors on our own
 		if (@\fwrite($this->writeHandle, $request) === false) {
 			throw new InitException(
 				"Writing to \"{$this->host}:{$this->port}}\" failed"
--- a/lib/Scanner/ICAPClient.php
+++ b/lib/Scanner/ICAPClient.php
@@ -25,13 +25,7 @@ class ICAPClient {
 
 		if (!$this->writeHandle) {
 			throw new InitException(
-				\sprintf(
-					'Cannot connect to "tcp://%s:%s": %s (code %d)',
-					$this->host,
-					$this->port,
-					$errorMessage,
-					$errorCode
-				)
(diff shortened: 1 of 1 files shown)


lib/Scanner/ICAPClient.php
```diff
@@ -15,28 +15,29 @@ public function __construct(string $host, int $port) {
...
+			5
+		);
+
+		if (!$this->writeHandle) {
+			throw new InitException(
+				\sprintf(
+					'Cannot connect to "tcp://%s:%s": %s (code %d)',
+					$this->host,
+					$this->port,
+					$errorMessage,
```
jvillafanez, line 32: Did you check it works properly? I'm not sure about the scope of the variables so it might always be null...

phil-davis:
Note: in PR #432 I ran the unit tests a lot of times with this code. They never failed. So somehow the changes here help/fix issue #420 without needing to increase the timeout that is now set using `stream_set_timeout`

lib/Scanner/ICAPClient.php
VicDeo: returned by `stream_socket_client` in case of failure:
errno Will be set to the system level error number if connection fails.
errstr Will be set to the system level error message if the connection fails.

https://www.php.net/manual/en/function.stream-socket-client.php

lib/Scanner/ICAPClient.php
```diff
@@ -15,28 +15,29 @@ public function __construct(string $host, int $port) {
...
 	}
 
 	private function disconnect(): void {
-		@\socket_shutdown($this->socket);
-		@\socket_close($this->socket);
-	}
-
-	public function getLastSocketError(): int {
-		return \socket_last_error($this->socket);
+		@\fclose($this->writeHandle);
```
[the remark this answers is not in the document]
VicDeo: Not errors but warnings. This just bloat a log file with useless info. I don't see any practical usage of the fact the socket was closed with a warning in case we got a proper answer from the daemon.
And even the answer from the daemon is cut and broken - we handle that in our code. 
So the success of operation has no matter.

lib/Scanner/ICAPClient.php
```diff
@@ -121,14 +122,18 @@ public function reqmod(string $service, array $body = [], array $headers = []):
 	}
 
 	private function send(string $request): string {
-		if (!$this->connect()) {
-			throw new RuntimeException("Cannot connect to icap://{$this->host}:{$this->port} (Socket error: " . $this->getLastSocketError() . ")");
+		$this->connect();
+		if (@\fwrite($this->writeHandle, $request) === false) {
```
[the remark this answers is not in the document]
VicDeo: just see  that we thrown exception just a line below so nothing is hidden in this particular case. 
 I'd say PHP errors are dumb for the whole set of these functions.

lib/Scanner/ICAPClient.php
jvillafanez: Normally you do something like:
```
$errorMessage = null;
stream_socket_client(....., $errorMessage);
echo $errorMessage;
```
The scope of the variable in this case is pretty clear.
However, in your case, I'm not sure if the scope of the `$errorMessage` will be limited to the `stream_socket_client` function, which would mean that the `$errorMessage` variable you're referring here is a new one and won't contain the error message of the `stream_socket_client`

In addition, for me, in the `stream_socket_client` function you're "using" the variable (as a value holder, I know), so I could think that the variable is defined / assigned elsewhere with a different value, so I wouldn't get an undefined variable error if I delete that function and change it for a different one.

lib/Scanner/ICAPClient.php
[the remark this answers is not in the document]
jvillafanez: Although it could be useless info, it would be something we should investigate at some point. No error in the log implies there is no bug, and if the app misbehaves for whatever reason we'll have problems if we need to investigate this area because we'll assume everything is ok when it might not.

Even a dumb warning such as "you're closing an already closed resource" could be a hint to detect potential performance problems above this code since we could be executing expensive operations twice or more times to reach here.
A comment would help to acknowledge that we're closing the resource multiple times (if that's the case), but if I don't see a comment saying that the server could have already closed the connection and we can get an error (I'm making up this, no idea if this could happen) I could assume this hasn't been taken care and that's something I could take into consideration if I need to debug this.

lib/Scanner/ICAPClient.php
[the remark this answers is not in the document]
jvillafanez: The same policy I have with "if you want use `==` write a comment explaining why not `===` instead", could be applied here: if you want to ignore a warning, explain why. An exception is thrown below, which is nice, but it doesn't explain why we couldn't write. If PHP just give dumb warnings, blame PHP for that, but that's what we have to work with.

jvillafanez:
@VicDeo https://github.com/owncloud/files_antivirus/pull/428#discussion_r622886661 is the only bit that concerns me. I'm not going to insist on the other 2 since they're mostly nice to have, but I don't expect too many problems there.

lib/Scanner/ICAPClient.php
VicDeo: ok, I swear it works  properly. :crossed_fingers: 
I guess I should send your points about the scope to PHP authors... ;)
as for me the code pretty clear: if I got `false` while opening the socket - I log `errcode` and `errnumber` to debug the reason. If I got anything that is non-false - the socket was opened properly and this is exactly what is written in docs for `stream_socket_client`:
>On success a stream resource is returned which may be used together with the other file functions (such as fgets(), fgetss(), fwrite(), fclose(), and feof()), false on failure.
errno
Will be set to the system level error number if connection fails.
errstr
Will be set to the system level error message if the connection fails.
so I guess the sane usage will always be  as per "Examples" section in https://www.php.net/manual/en/function.stream-socket-client.php
```
$fp = stream_socket_client("tcp://www.example.com:80", $errno, $errstr, 30);
if (!$fp) {
    echo "$errstr ($errno)<br />\n";
} else {
    fwrite($fp, "GET / HTTP/1.0\r\nHost: www.example.com\r\nAccept: */*\r\n\r\n");
    while (!feof($fp)) {
        echo fgets($fp, 1024);
    }
    fclose($fp);
}
```
Also I don't see a point in deletion of random code pieces... Let's delete random letters in code - it's even more funny. :)

VicDeo: I mean that the fact `!$fp` is true always mean both `$errstr ($errno)` are defined. This is how PHP authors suggest to process errors for such functions.

jvillafanez: "You win this time" :stuck_out_tongue: 
The code is small enough to see the whole function, so if we ever need to touch this function I don't think we'll make dumb mistakes.

lib/Scanner/ICAPClient.php
```diff
@@ -15,28 +15,29 @@ public function __construct(string $host, int $port) {
...
+		$this->writeHandle = @\stream_socket_client(
+			"tcp://{$this->host}:{$this->port}",
+			$errorCode,
+			$errorMessage,
+			5
+		);
+
+		if (!$this->writeHandle) {
+			throw new InitException(
+				\sprintf(
```
jvillafanez, line 28: String interpolation is preferred

lib/Scanner/ICAPClient.php
```diff
@@ -121,14 +122,18 @@ public function reqmod(string $service, array $body = [], array $headers = []):
 	}
 
 	private function send(string $request): string {
-		if (!$this->connect()) {
-			throw new RuntimeException("Cannot connect to icap://{$this->host}:{$this->port} (Socket error: " . $this->getLastSocketError() . ")");
+		$this->connect();
+		if (@\fwrite($this->writeHandle, $request) === false) {
+			throw new InitException(
+				\sprintf('Writing to "%s:%s" failed', $this->host, $this->port)
```
jvillafanez, line 128: string interpolation is preferred

jvillafanez (approved):
Just 2 tiny details, but I'm not blocking. Up to you.

lib/Scanner/ICAPClient.php
VicDeo: Ok... but it will be jerky due to a mixture of single/double quotes. I need double quotes inside the message itself so
`sprintf` looks more clean.

jvillafanez: you could remove the double quotes, or change them to single quotes. Taking into account that the neither the host or port will have a space, I doubt people will misinterpret the tcp string as multiple strings.

VicDeo: nope... it's double quotes are expected here. E.g. https://github.com/owncloud/files_antivirus/blob/master/lib/Scanner/Socket.php#L60

VicDeo: I had used a clean way to concat args in this repo years before it was decided that "the interpolation is preferred" :bearded_person:

VicDeo: By the way the complete phrase sounds like "Interpolation is preferred to concatenation". And `sprintf` is NOT a concatenation - it's still interpolation. Anyway, you wanted it dirty - you got it dirty ;) 
Personally I dislike my last commit.

VicDeo:
@jvillafanez Updated. Could you please rereview?

jvillafanez (approved):

phil-davis:
> Note: in PR #432 I ran the unit tests a lot of times with this code. They never failed. So somehow the changes here help/fix issue #420 without needing to increase the timeout that is now set using `stream_set_timeout`

The previous run of CI  for this PR - https://drone.owncloud.com/owncloud/files_antivirus/1882/5/9 - had `RuntimeException: Unknown ICAP response` - so maybe there is not "magic" here that fixes issue #420 

I restarted CI.

PR #430, opened 2021-04-29
Increase socket timeout from 5 to 10 seconds

phil-davis:
It might help issue #420


phil-davis:
I did lots of runs of the unit tests. All unit test runs passed - great.
So it is the socket timeout that causes intermittent unit test fails.

phil-davis:
Note: also see PR #432 - the coming change to use stream socket client seems to also fix the intermittent tests without needing to increase the timeout.

PR #431, opened 2021-04-29
Reduce socket timeout to demonstrate issue 420

phil-davis:


phil-davis:
Random fails - for example https://drone.owncloud.com/owncloud/files_antivirus/1852/9/8
`RuntimeException: Unknown ICAP response`

That is what is expected when the timeout is too low.

PR #432, opened 2021-04-29
[tests-only] Trigger ci for stream socket client

phil-davis:
To see if we get intermittent fails


phil-davis:
No intermittent fails of unit tests. That is good (or lucky?)

PR #434, opened 2021-05-05
Check content before create for public upload case

### LINKED ISSUE #334: Can upload virus file with new public WebDAV API
### Steps to reproduce
1. Setup files antivirus on owncloud instance.
2. Enable `tech_preview` in order to use new public WebDAV API.
```
php occ config:system:set dav.enable.tech_preview --value true --type bool
```
3. Create a password protected public link share.
4. Try to upload a virus file with new public WebDAV API.
eg.
```
curl http://public:1234@172.17.0.1/oc/core/remote.php/dav/public-files/FkQv9oZCKMIbmfl/eicar.com -X PUT -H "Content-Type: text/plain" --data 'X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*'
```

### Expected Behavior
The file will not be uploaded to public share

### Actual Result
The file gets uploaded and is in locked state.

VicDeo:
In case of public upload we need to scan the content before the file is created. 
Otherwise after the virus is detected we end up with an empty locked file.

Fixes #334

### DIFF
--- a/tests/acceptance/features/apiAntivirus/antivirusMain.feature
+++ b/tests/acceptance/features/apiAntivirus/antivirusMain.feature
@@ -121,10 +121,8 @@ Feature: Antivirus basic
       | eicar_com.zip   |
       | eicarcom2.zip   |
 
-  @skip @issue-334
   Scenario Outline: A small file with a virus cannot be uploaded via new public upload
-    Given the administrator has enabled DAV tech_preview
-    And as user "Alice"
+    Given as user "Alice"
     And user "Alice" has created a public link share of folder "FOLDER" with change permissions
     When the public uploads file "<virus-file-name>" from the antivirus test data folder using the new WebDAV API
     Then the HTTP status code should be "403"
--- a/lib/AppConfig.php
+++ b/lib/AppConfig.php
@@ -218,7 +218,11 @@ class AppConfig {
 	 */
 	protected function getter($key) {
 		if (\array_key_exists($key, $this->defaults)) {
-			return $this->getAppValue($key);
+			$value = $this->getAppValue($key);
+			if ($key === 'av_max_file_size') {
+				return (int) $value;
+			}
+			return $value;
 		} else {
 			throw new \BadFunctionCallException($key . ' is not a valid key');
 		}
--- a/lib/Dav/AntivirusPlugin.php
+++ b/lib/Dav/AntivirusPlugin.php
@@ -7,7 +7,7 @@
  *
  * @author Viktar Dubiniuk <dubiniuk@owncloud.com>
  *
- * @copyright Viktar Dubiniuk 2018
+ * @copyright Viktar Dubiniuk 2021
  * @license AGPL-3.0
  */
 
@@ -15,6 +15,10 @@ namespace OCA\Files_Antivirus\Dav;
 
 use OCA\DAV\Upload\FutureFile;
 use OCA\Files_Antivirus\AppInfo\Application;
+use OCA\Files_Antivirus\Resource;
+use OCA\Files_Antivirus\Status;
+use Sabre\DAV\Exception\Forbidden;
+use Sabre\DAV\INode;
 use Sabre\DAV\Server;
 use Sabre\DAV\ServerPlugin;
 
@@ -27,7 +31,7 @@ class AntivirusPlugin extends ServerPlugin {
 	/**
 	 * @var \Sabre\DAV\Server $server
 	 */
-	private $server;
+	private $davServer;
 
 	/**
 	 * @var Application
@@ -56,8 +60,9 @@ class AntivirusPlugin extends ServerPlugin {
 	 * @return void
 	 */
 	public function initialize(Server $server) {
-		$this->server = $server;
-		$this->server->on('beforeMove', [$this, 'beforeMove'], 1);
+		$this->davServer = $server;
+		$this->davServer->on('beforeMove', [$this, 'beforeMove'], 1);
+		$this->davServer->on('beforeCreateFile', [$this, 'beforeCreateFile']);
 	}
 
 	/**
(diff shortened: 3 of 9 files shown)


tests/acceptance/features/apiAntivirus/antivirusMain.feature
```diff
@@ -121,7 +121,6 @@ Feature: Antivirus basic
       | eicar_com.zip   |
       | eicarcom2.zip   |
 
-  @skip @issue-334
   Scenario Outline: A small file with a virus cannot be uploaded via new public upload
     Given the administrator has enabled DAV tech_preview
```
VicDeo, line 125: @phil-davis I think I should delete this line too as tech preview parameter  is not needed anymore.

phil-davis: @VicDeo yes, any reference to tech_preview can be removed.

phil-davis:
https://drone.owncloud.com/owncloud/files_antivirus/1905/17/12
```
  Scenario Outline: Files bigger than the upload threshold are not checked for viruses when uploaded via public upload        # /var/www/owncloud/testrunner/apps/files_antivirus/tests/acceptance/features/apiAntivirus/antivirusFileSize.feature:118
    Given the administrator has enabled DAV tech_preview                                                                      # OccContext::theAdministratorHasEnabledDAVTechPreview()
    And as user "Alice"                                                                                                       # FeatureContext::asUser()
    And user "Alice" has created a public link share of folder "FOLDER" with change permissions                               # FeatureContext::userHasCreatedAPublicLinkShareOfWithPermission()
    And parameter "av_max_file_size" of app "files_antivirus" has been set to "100"                                           # AppConfigurationContext::serverParameterHasBeenSetTo()
    When the public uploads file "eicar_com.zip" from the antivirus test data folder using the <public-webdav-api> WebDAV API # AntivirusContext::publicUploadsFileFromAntivirusDataFolder()
    Then the HTTP status code should be "201"                                                                                 # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    And as "Alice" file "/FOLDER/eicar_com.zip" should exist                                                                  # FeatureContext::asFileOrFolderShouldExist()

    Examples:
      | public-webdav-api |
      | new               |
        HTTP status code 403 is not the expected value 201
        Failed asserting that 403 matches expected '201'.
      | old               |

  Scenario Outline: Files bigger than the upload threshold are not checked for viruses when uploaded overwriting via public upload                             # /var/www/owncloud/testrunner/apps/files_antivirus/tests/acceptance/features/apiAntivirus/antivirusFileSize.feature:144
    Given the administrator has enabled DAV tech_preview                                                                                                       # OccContext::theAdministratorHasEnabledDAVTechPreview()
    And as user "Alice"                                                                                                                                        # FeatureContext::asUser()
    And user "Alice" has created a public link share of folder "FOLDER" with change permissions                                                                # FeatureContext::userHasCreatedAPublicLinkShareOfWithPermission()
    And parameter "av_max_file_size" of app "files_antivirus" has been set to "60"                                                                             # AppConfigurationContext::serverParameterHasBeenSetTo()
    When the public uploads file "textfile.txt" from the antivirus test data folder using the <public-webdav-api> WebDAV API                                   # AntivirusContext::publicUploadsFileFromAntivirusDataFolder()
    And the public overwrites file "textfile.txt" with content "X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*" using the old WebDAV API # PublicWebDavContext::thePublicOverwritesFileWithContentUsingOldWebDavApi()
    Then the HTTP status code should be "204"                                                                                                                  # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    And the content of file "/FOLDER/textfile.txt" for user "Alice" should be "X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*"           # FeatureContext::contentOfFileForUserShouldBe()

    Examples:
      | public-webdav-api |
      | new               |
      | old               |
```

Only the case of a large upload to the new public-webdav-api is failing. It passes when using the old public-webdav-api - so I guess that the new code in this PR is not being triggered for the old public-webdav-api.

And in the 2nd scenario outline, both cases pass. The antivirus check does not seem to be happening when the "large" public upload overwrites an existing file.

Also, at the moment, the antivirus app is designed so that files above a certain size are not checked for viruses. It is a requirements decision (product manager?) if public uploads should always be checked, or if the size limit should also apply to public uploads.

VicDeo:
@phil-davis imo the file  limit should always be applied but in this particular case I have no technical ability to get the size before the scan is started as I'm reading from the resource (a file handle) and this resource doesn't report the file size.

phil-davis:
> @phil-davis imo the file limit should always be applied but in this particular case I have no technical ability to get the size before the scan is started as I'm reading from the resource (a file handle) and this resource doesn't report the file size.

Maybe it is acceptable that size limits do not apply to public uploads? It is more secure to always scan every upload. But also, admins might have set the size limit because of some resource limit on the server+scanner.

@pmaier1 is this a requirement that you can decide about?

VicDeo:
@C0rby 
> IMHO the failing scenario is correct like this now. Or was it intended that you could upload a virus if it is bigger then the max size?

```
man clamd.conf
       LogFileMaxSize SIZE
              Maximum size of the log file.
              Value of 0 disables the limit.
              Default: 1048576
...
NOTES
       All options expressing a size are limited to max 4GB. Values in excess will be reset to the maximum.
```
https://security.stackexchange.com/questions/107132/linux-antivirus-and-files-bigger-than-4gb

C0rby:
@VicDeo, ah good to know. That is new to me...

I think I also misunderstood the test case.... It probably means that if a file is bigger then the max size it wouldn't even be allowed to be uploaded so we don't need to scan it, right?

Then that should be fine I guess.

VicDeo:
@C0rby not exactly. clamav has a hard filesize limit - 4Gb.  E.g. scanning a blue ray image could be too resource consuming.
But we introduced  two options to overcome that in the app:
`File size limit` and `Stream Length`
So it is possible to set 
`File size limit` to -1 (no limit)
and `Stream Length` to e.g. 2Gb
In this case it is possible to scan any size but it will take even more resources:
the file will be cut into  2Gb chunks with some internal overlapping and every chunk will need a separate request to clamav. All requests are required to be successful to consider the whole file clean.

That's why I think we need to respect `File size limit` and `Stream Length` for the anonymous upload too.

On the other hand: if the virus is present but the file size is over the `File size limit` - the file should **NOT** be scanned (so the virus is undetected)
And exactly this type of tests we have failing here.

VicDeo:
@phil-davis ok, The reason is very funny as the limit of 100 bytes is rarely used in the real life: 
we read from the stream in chunks and the chunk size is 8 kb. So it's kind of a corner case - I just need to trim the content before it is sent to the clamav if the "length of the_last_chunk + bytes_sent_before" is over  `av_max_file_size`  to exactly match `av_max_file_size`
Should be fixed now.

phil-davis:
The limit is set low in the tests just to make it easy to create files either side of the limit and test the behavior. If you think it would be better to use a more realistic limit, we can adjust the test suite.

VicDeo:
@phil-davis 
>If you think it would be better to use a more realistic limit, we can adjust the test suite.

no, I'm absolutely fine with the current limit. it's just a fun fact.

VicDeo:
it's killing me...
```
>  docker run -ti deepdiver/icap-clamav-service -p 1344:1344

LibClamAV Warning: **************************************************
LibClamAV Warning: ***  The virus database is older than 7 days!  ***
LibClamAV Warning: ***   Please update it as soon as possible.    ***
LibClamAV Warning: **************************************************
recomputing istag ...
recomputing istag ...
```
and I have no clue why 127.0.0.1:1344 refuses to connect :/

VicDeo:
finally fine...

phil-davis:
Those annoying randm unit test fails are back: https://drone.owncloud.com/owncloud/files_antivirus/1954/4/8
```
PHPUnit 8.5.15 by Sebastian Bergmann and contributors.

Runtime:       PHPDBG 7.2.34-21+ubuntu18.04.1+deb.sury.org+1
Configuration: /var/www/owncloud/server/apps/files_antivirus/phpunit.xml

......................E.........                                  32 / 32 (100%)

Time: 14.53 seconds, Memory: 56.52 MB

There was 1 error:

1) OCA\Files_Antivirus\Tests\unit\Scanner\IcapScannerTest::testScannerAsync with data set "eicar" (1, 'X5O!P%@AP[4\PZX54(P^)7CC)7}$E...!$H+H*')
RuntimeException: Unknown ICAP response: ""

/var/www/owncloud/server/apps/files_antivirus/lib/Scanner/ICAPClient.php:151
/var/www/owncloud/server/apps/files_antivirus/lib/Scanner/ICAPClient.php:116
/var/www/owncloud/server/apps/files_antivirus/lib/Scanner/ICAPScanner.php:54
/var/www/owncloud/server/apps/files_antivirus/tests/unit/Scanner/IcapScannerTest.php:55

ERRORS!
Tests: 32, Assertions: 52, Errors: 1.
```
I will restart drone CI.

phil-davis (commented):
CI is passing now. @jvillafanez review please.

jvillafanez (approved):
I don't really like the `AntivirusPlugin` to have a dependency with the whole application... something for a different PR I guess.

PR #435, opened 2021-05-07
Increase stream_set_timeout from 5 to 10 seconds

phil-davis:
So that we have less chance of getting timeouts.

Issue #420 - we want to make the CI tests a bit more robust, and this should also be fine for production use.


phil-davis:
PR #436 demonstrates running the unit tests another 11 times - they all passed.

Increasing the timeout from 5 to 10 seconds seems to help CI be reliable.

@VicDeo @jvillafanez does this seem OK? Is there any reason not to do this? Please review.

jvillafanez (approved):

PR #436, opened 2021-05-07
[tests-only] Test 435

phil-davis:
Just run the unit tests in CI, without the Oracle pipeline (it is so slow!), with the timeout change from PR #435


phil-davis:
12 runs of CI - all unit tests passed. That is "a good thing" (tm)

phil-davis:
PR #435 has been merged - good.

Issue #438, opened 2021-05-12
[QA] Error 500  "AV failed!" appears if to upload empty file using ICAP connected to Kaspersky Scan engine


martjanoff:
**Pre-condition:**
ICAP connected to Kaspersky Scan engine is set up on machine.

**Steps to reproduce:**
- Log in as admin
- Go to /owncloud/index.php/apps/files
- Upload empty file ( tested with .txt, .zip, .doc)

**Expected result:**
Should be able to upload the file.

**Actual result:**
Error 500  "AV failed!" appears.

![image](https://user-images.githubusercontent.com/37621174/117992458-9fcd1680-b347-11eb-989f-1117e5541dc5.png)

phil-davis:
I added this case to the automated tests - PR #439 
That is passing. The test suite uses `deepdiver/icap-clamav-service` as the ICAP service.

VicDeo:
To be validated in the next Beta/RC

jnweiger:
Confirmed fixed in v1.0.0-rc2

PR #439, opened 2021-05-12
[tests-only] Test upload of empty file

phil-davis:
Add a test scenario for uploading an empty file, to check issue #438 

We may as well cover this edge case with a test scenario.


VicDeo:
inter-resting... 
https://github.com/owncloud/files_antivirus/blob/b01b8691a5d5f5ef17bab33daa78b3ee18088c16/lib/Scanner/ICAPScanner.php#L37-L39

phil-davis:
> according to the code it should not ask to scan empty content but bail out with success immediately

which means that there should not be any problem, no matter what ICAP service is being used.

VicDeo:
ok... I need to check. There is  a chance `$this->data === ''` is `false` due to `$this->data`  is null.
strict comparison as is ;)

phil-davis:
> ok... I need to check

It would be nice if a unit test covered that. But it might be tricky because the method will have the same result regardless of which code path is taken.

individual-it (approved):

PR #444, opened 2021-05-21
Hotfix for slow virus scanner responses

jnweiger:
This implements the hotfix suggsted in #443

PR #445, opened 2021-05-25
[ICAP] Stop reading the response after headers are read

### LINKED ISSUE #443: [QA] cannot upload files larger than ca. 20 MB
Seen with files_antivirus 1.0.0-rc2 running on https://oc1070-macafee-20210519.jw-qa.owncloud.works/owncloud connected to
   Ralf's McAfee ICAP instance. Both ownCloud and McAfee are on separate hetzner machines. Thus file transfer times matter.

   * 1.0.0Beta1 had no read timeout (only ClamAV and Kaspersky were supported)
   * 1.0.0-beta2 introduced a 5 sec timeout. (McAfee support was added)
   * 1.0.0-rc1 and 1.0.0-rc2 have a 10 sec timeout.

   When the timeout occurs ownCloud reports `Unknown ICAP response: ""` and blocks the upload.
   An exe file of ca 18 MB sporadically hits the limit. Bigger files (33, 41, 81, 125 MB as e.g. listed below) hit the limit reliably.
   https://download.owncloud.com/desktop/ownCloud/stable/latest/win/            (ca 18 MB)
   https://inkscape.org/gallery/item/8396/Inkscape-0.48.5-1-win32.exe           (33 MB)
   https://inkscape.org/gallery/item/3944/Inkscape-0.91-1.exe                   (41 MB)
   https://inkscape.org/gallery/item/18066/inkscape-0.92.5-x64.msi              (82 MB)
   https://media.inkscape.org/dl/resources/file/inkscape-1.0.2-2-x64.msi        (125 MB)

   The timeout was introduced for McAfee, but also affects ClamAV and Kasperky.

   Bad workaround:
        * increase the timeout to XXX sec.  (150 sec is sufficient for up to 82 MB in the above setup)
        * this makes us wait XXX sec on each(!) upload (even very small files)
   Acceptable hotfix:
        * increase the timeout on the very first read to 300 sec, and on all subsequent reads lower to 2 sec. This reliably proesses
          processes all these files, and takes 5 times faster on small files, than the current implementation.
   ICAP compliant implementation:
        * Parse the Content-Length: header in the response envelope, read exactly this payload length. No timeouts needed.

VicDeo:
Do not read the whole response as we need only ICAP headers and `Encapsulated` header.

So we read ICAP headers one by one until `Encapsulated` header is met.
Then we parse offsets provided by `res-hdr` and `res-body` of this header.

the content is considered to be **infected** if ANY of the following is true:
- ICAP headers contain the header specified in `av_response_header` 
- HTTP header  in `res-hdr` section has status code `403`

the content is considered to be **clean** if the ICAP status code is `204`

### TL;DR
McAfee just hangs so  if we want to read whole response we should wait until the network timeout happens
But the whole response is not needed for the virus detection as
- 204 ICAP status means 'clean'
- 200 ICAP status with encapsulated 403 HTTP status means 'infected'

Fixes #443 
Obsoletes https://github.com/owncloud/files_antivirus/pull/444


VicDeo:
@jnweiger if it possible to test this patch be fore merge?

jnweiger:
> @jnweiger if it possible to test this patch before merge?

OK. I'll give it a test first, then we merge.

jnweiger (changes requested):
With this patch, an infected file is reported immediately.
A a clean file hangs for 60 seconds in 'processing file'.
Seen with McAfee via ICAP.

ClamAV and Kaspersky via ICAP work fine.

VicDeo:
@jnweiger 
> With this patch, an infected file is reported immediately.
A clean file hangs for 60 seconds in 'processing file'.
Seen with McAfee via ICAP.

So is it acceptable or not?
If not I can't go further without access to McAfee.

jnweiger:
> @jnweiger
> 
> > With this patch, an infected file is reported immediately.
> > A clean file hangs for 60 seconds in 'processing file'.
> > Seen with McAfee via ICAP.
> 
> So is it acceptable or not?
> If not I can't go further without access to McAfee.

A one minute delay per normal file is not acceptable. @VicDeo You now have access to my test system.

VicDeo:
@JanAckermann could you please sign the CLA for this repo?

JanAckermann:
Done

JanAckermann (approved):
Codewise LGTM

PR #450, opened 2021-06-11
Scan public upload on update

VicDeo:
Fixes https://github.com/owncloud/files_antivirus/issues/334 completely

Now when a new content of the file in the public upload folder is infected we just keep the old content.
The behavior spotted in https://github.com/owncloud/files_antivirus/issues/334#issuecomment-848609138 (a stray DB entry with a missing file on the FS) is no longer reproducible.

```
> URL="http://localhost/~deo/oc-tmp/remote.php/dav/public-files/jfsT7ujGSK8JMNn/somenewfile.txt" 
deo@jah-mobile:~> curl $URL -X PUT --data "Hello World"                                        
deo@jah-mobile:~> curl $URL -T ~/public_html/_craft/testfiles/eicar.txt
<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>OCP\Files\FileContentNotAllowedException</s:exception>
  <s:message>Virus Type=0; Resolution=2; Threat=Win.Test.EICAR_HDB-1; is detected in the file. Upload cannot be completed.</s:message>
</d:error>
```

### DIFF
--- a/lib/Dav/AntivirusPlugin.php
+++ b/lib/Dav/AntivirusPlugin.php
@@ -17,6 +17,8 @@ use OCA\DAV\Upload\FutureFile;
 use OCA\Files_Antivirus\AppInfo\Application;
 use OCA\Files_Antivirus\Resource;
 use OCA\Files_Antivirus\Status;
+use OCP\ILogger;
+use OCP\IUserSession;
 use Sabre\DAV\Exception\Forbidden;
 use Sabre\DAV\INode;
 use Sabre\DAV\Server;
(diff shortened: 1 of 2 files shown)


lib/Dav/AntivirusPlugin.php
```diff
@@ -90,31 +105,61 @@ public function beforeMove($source, $destination) {
...
+	 * @param INode $node
+	 * @param resource $data
+	 * @param bool $modified should be set to true, if this event handler
+	 *                           changed &$data
+	 * @return bool|null
+	 * @throws Forbidden
+	 */
+	public function beforeWriteContent($path, INode $node, &$data, &$modified) {
+		// Scan a public upload
+		if ($this->userSession->getUser() === null) {
```
micbar, line 133: this check was not present before. Is this codepath really only for public uploads?

VicDeo: the whole method `beforeWriteContent` is a brand new one. 
as for `beforeCreateFile`  method  - this check existed there: https://github.com/owncloud/files_antivirus/blob/3873ea322230fd0544e545839703ad5def7a2b0c/lib/Dav/AntivirusPlugin.php#L98

micbar (approved):

PR #474, opened 2022-02-17
[tests-only] Start federated server for core-cli test pipelines

phil-davis:
Issue https://github.com/owncloud/core/issues/39799

PR #475, opened 2022-02-17
[tests-only] Test core cli federated server

phil-davis:
test the change in PR #474 by running the core-cli pipelines in this PR (rather than only in nightly CI)


phil-davis:
`core-cli` pipelines have run and passed - good.

PR #481, opened 2022-03-18
[CI] Updated drone.star file to match latest changes

grgprarup:
## Description
This is an update on drone.star file 
- using drone-cancel-previous-builds
- using thegeeklab/drone-github-comment instead of jmccann/drone-github-comment for github comments 
- moving docker image references to constants
- using caches images
- Do not run PHP unit tests with Oracle by default
- Skip pipelines based on changed code

## Related Issue
Part of: 
- https://github.com/owncloud/QA/issues/700
- https://github.com/owncloud/QA/issues/717
- https://github.com/owncloud/QA/issues/729
- https://github.com/owncloud/QA/issues/728
- https://github.com/owncloud/core/issues/38679
- https://github.com/owncloud/QA/issues/730

PR #482, opened 2022-03-18
[tests-only] enable files_external app when running core tests

phil-davis:
because some core tests are testing with local_storage and that requires the files_external app to be enabled.


phil-davis:
done in #481

PR #514, opened 2022-11-16
fix: an upload cannot be deleted but only denied + use the same message in two places to reduce translation effort

DeepDiver1975:
to be merged after #512


jnweiger:
Confirmed fixed in 1.2.0-rc.1

Issue #501, opened 2022-08-08
[QA] duplicate UI elements when switching from fortinet to mcafee


jnweiger:
Seen with files_antivirus 1.1.0-rc.3

* In the admin web interface, Security, files_antivirus select
  - select Mode `Daemon (ICAP)`
  - then select Mode `Fortinet (ICAP)` - the fields and descriptions change with some fancy animation to match the new mode.
  - then select Mode `McAfee Webgateway 10.x and higher (ICAP)` - again, the UI changes with an animation
  - then select Mode `Fortinet (ICAP)` - the existing *ICAP request service* and *ICAP response header* fields scroll down, and two new copies appear nicely animated. BAD
  
![image](https://user-images.githubusercontent.com/1108546/183461634-a1853ff5-e476-4873-83b2-2c31aac80c1f.png)

* Functionality is not affected by this UI glitch. The duplicates are inactive, the new fields work as expected.
* Only the transition from `McAfee Webgateway 10.x and higher (ICAP)` to `Fortinet (ICAP)` is affected, all other transitions look correct.

Expected behaviour
* Fields change inplace, when they have to. No inactive duplicates appear.

jnweiger:
Confirmed fixed in 1.2.0-rc.1

PR #516, opened 2022-11-18
fix: av mode names & fortinet fields

DeepDiver1975:
Screenshot from 2022-11-18 08-34-06
Screenshot from 2022-11-18 08-33-56


DeepDiver1975:
fixes #501

jnweiger (approved):
Thanks!

PR #528, opened 2022-11-22
[tests-only] Check PHP code with 7.3

phil-davis:
oC10 core up to 10.11 supports PHP 7.3. For the time being, when we release core apps, we want to know that the PHP code works OK with PHP 7.3. So run code-style in CI with PHP 7.3

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -50,11 +50,19 @@ config = {
         "master",
     ],
     "appInstallCommandPhp": "make",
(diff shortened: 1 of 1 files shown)


phil-davis:
https://drone.owncloud.com/owncloud/files_antivirus/2783/2/3
coding-standard-php7.3 passes - good.

individual-it (approved):

PR #535, opened 2022-11-28
[tests-only] [full-ci] Run phan with PHP 7.3

phil-davis:
`phan` will report errors if any of the code is not compatible with PHP 7.3.

This app still supports PHP 7.3 when used with oC10 core 10.11 and earlier.

To get `phan` passing, I suppressed most of the things it reported. They all look like OK code to me. There are a few deprecated constants, but I don't want to change them to their newer names, because that might break if running with some older core version with older dependenices or...


phil-davis:
https://drone.owncloud.com/owncloud/files_antivirus/2802/5/3
`phan` passed with PHP 7.3 (and 7.4) - good

.phan/config.php
```diff
@@ -0,0 +1,111 @@
+<?php
+/**
+ * @author Patrick Jahns <github@patrickjahns.de>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
```
kiranparajuli589, line 5:
```suggestion
 * @copyright Copyright (c) 2022, ownCloud GmbH
```

.phan/config.php
```diff
@@ -0,0 +1,111 @@
+<?php
+/**
+ * @author Patrick Jahns <github@patrickjahns.de>
```
kiranparajuli589, line 3:
```suggestion
 * @author Patrick Jahns <github@patrickjahns.de>
```
:arrow_up:

phil-davis: This is a copy-paste of the config from other repos, originally written by Patrick. I don't know what the rules are when doing that sort of thing. In the other repos, I have been leaving his name.

kiranparajuli589 (approved):

PR #538, opened 2023-04-03
fix: don't use `Preview` with McAffee - this will result in a `100` r…

DeepDiver1975:
…esponse code, which we never implemented properly

## Background

In preview mode McAffee is responding with status code 100 which results in McAffee to wait for more data to be sent by ownCloud. But we have not more data to send - and `100` handling was never properly implemented.

As a result `Preview` is removed from the McAffee implementation of ICAP scanner & client. 

fixes https://github.com/owncloud/enterprise/issues/5588

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased] - 2022-xx-xx
 
--
+### Fixed
+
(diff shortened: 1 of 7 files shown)


jnweiger (approved):
Tested with my local mcafee setup: positive and negative behaviour is correct. No 60sec delay any more.

phil-davis (approved):
Looks OK to me

jnweiger:
Released as part of v1.2.2

PR #540, opened 2023-05-13
Fix hostname input validation

### LINKED ISSUE #523: [QA] Invalid IP-Address should be rejected
Seen in 1.2.0-rc.1

According to https://github.com/owncloud/QA/edit/master/Server/Test_Plan_Files_Antivirus.md
* The given example  'localhost%&$#+~' of an invalid hostname is rejected. OK
grafik
(wait, ... the UI is in English, why does this popup talk German?)

* An invalid ip address, e.g. '127.0.0.1.1' is not rejected. BAD
grafik

* The test scan during save reports an error. OK

Expected behaviour:
* both, invalid hostname and IP-Address are handled the same.
* (and after swithcing from German to English + SHIFT-RELOAD in the browser, no German artefacts remain.)

prsnbrg:
Fix #523 

Tested with different strings:
- 127.0.0.1 -> valid
- 256.0.0.1 -> invalid
- 127.0.0.1.1 -> invalid
- localhost -> valid
- localhost.me -> valid
- localhost"!? -> invalid


phil-davis:
Other wacky things:
- abc.x - invalid (the last "x" part must be at least 2 characters long)
- 1111.com - valid (and yes, I think that the hostname spec allows components of the hostname to be all dots)
- 1234 - valid (this is like "localhost" or other local names, and I think all-digits is allowed (although it will be confusing if admins do that - define local hostnames that are all digits)

This works for me - I can't find anything wrong so far.

phil-davis:
Note: we cannot enter an IPv6 address like `2001:0db8:85a3:0000:0000:8a2e:0370:7334`
But that was not possible before this PR either. IPv6 addresses have colons in them (and there are multiple formats that they can be written in. So that is a separate thing, if someone needs to input a literal IPv6 address. (usually I would expect a "normal" hostname that then might lookup in DNS an turn out to be IPv6 to reach that hostname)

phil-davis (approved):

phil-davis:
@jnweiger please review and merge if you are happy with this.

jnweiger (approved):
LGTM. Thanks!

PR #547, opened 2023-06-14
allow longer TLDs

jnweiger:
with max 6 characters we would exclude e.g. .international, .company, .university ...


jnweiger:
Note that this goes into release-1.2.2 branch, which will be merged back to master after the release.

phil-davis (approved):

PR #564, opened 2026-03-03
feat: oc11

DeepDiver1975:


phil-davis (approved):
Looks OK. Why do we need lots of`phpstan-ignore-line`?

